PR #4, opened 2018-04-11
PRO-2327 Added unit tests for ExecutorsAddressValidationRule

malinovsky:


sanjayparekh-hmcts:
Those changs are good
But we really need the same for DeceasedAddressValidationRule too

PR #182, opened 2018-11-13
Pro 2795 remove escaped characters in declarartion

sanjayparekh-hmcts:
PRO-2795: need to remove encoded chars in the legal statement + delclaration for PA applications when doing Print from ccd-ui
[ ] Yes
[X ] No
```


sanjayparekh-hmcts:
Dont need this change

PR #188, opened 2018-11-29
Pro 4108 caveator generic email

RababRabab:

PRO-4108 - Generic email to caveator

### Change description ###

Added the ability to trigger generic email messages from most of the states within the caveats workflow.

[ ] Yes
[x] No


src/main/java/uk/gov/hmcts/probate/model/ccd/caveat/request/CaveatData.java
```diff
@@ -2,18 +2,24 @@
...
 
 import java.time.LocalDate;
+import java.util.ArrayList;
 import java.util.List;
 
 @Builder
 @Data
 public class CaveatData {
 
+    private final ApplicationType cavApplicationType;
```
malinovsky, line 20: Since we are in Caveat specific class already, can we get rid of `cav` prefixes?

malinovsky (approved):

PR #222, opened 2019-01-10
PRO-4414 Extend grant templates to support registry addresses

RebeccaBaker:
https://tools.hmcts.net/jira/browse/PRO-4414

### Change description ###
Updated templates to support multiple registry addresses. Update unit and functional tests. Update cosmetic issues around the layout and where the hologram will sit. Updated to include additional clause about verifying the data.

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/probate/controller/DocumentController.java
```diff
@@ -90,11 +94,11 @@
     public ResponseEntity<CallbackResponse> generateGrant(@RequestBody CallbackRequest callbackRequest)
             throws NotificationClientException {
 
-        List<Document> documents = new ArrayList<>();
         CaseDetails caseDetails = callbackRequest.getCaseDetails();
         @Valid CaseData caseData = caseDetails.getData();
         DocumentType template;
         Document digitalGrantDocument;
+        getRegistryDetails(callbackRequest);
```
ikenna-hmcts: You could pass caseDetails as a parameter to getRegistries since we already have it.

src/main/java/uk/gov/hmcts/probate/model/ccd/raw/request/CaseDetails.java
```diff
@@ -20,4 +20,12 @@
     private final Long id;
 
     private String grantSignatureBase64;
+
+    private String registryAddressLine1;
```
ikenna-hmcts, line 24: So why are we not adding registry Address to caseData like solicitor and Pa instead of in caseDetails? Also wondering why you did not use A registryAddress object to hold the address details

RebeccaBaker: Can't use caseData, as the registry addresses are held in the application.yaml and not entered in by the user, so it is not stored against the case. The reason for this is so that the case can move from registry to registry without need to do if logic and update the case data.

src/main/java/uk/gov/hmcts/probate/controller/DocumentController.java
RebeccaBaker: Committed change to use caseDetails as opposed to callBackRequest

ikenna-hmcts (approved):

RebeccaBaker:
Closing due to Upgrade on jenkins build causing issues

PR #226, opened 2019-01-14
PRO-4483: Update to fix scannedDate not being seen as an ISO 8601 dat…

davejones74:
https://tools.hmcts.net/jira/browse/PRO-4483

### Change description ###
A number of errors was causing an issue when using event callbacks with backoffice code when scanned documents are involved. This particular PR should be the last of these issues and handles scannedDate field for a ScannedDocument which was recently changed from a String to a LocalDateTime field. This caused CCD not to recognise this as an ISO 8601 Date. Specifically this mean that it was missing the timezone or a 'Z' on the end of the string.

Also a small change was done to the tests to create an independant ScannedDocument check.

```
[ ] Yes
[x] No
```


dhruvpatel20:
@davejones74 A few things:
- As you are adding this to the main application configuration - will it have an impact on any existing flows?
- You will need to fix the conflicts on this PR
- Also the coverage has slipped very slightly - please could you increase it a bit? (otherwise the build fails)

davejones74:
> @davejones74 A few things:
> 
>     * As you are adding this to the main application configuration - will it have an impact on any existing flows?
I've check and only Notifications use LocalDateTime and that has it's own DateTimeFormatter applied so this is fine.
 
>     * You will need to fix the conflicts on this PR
Done
 
>     * Also the coverage has slipped very slightly - please could you increase it a bit? (otherwise the build fails)
I've split out the serialization into a seperate class which makes it easier to add tests for this, also added a new test for an unrelated method to improve overall code coverage.

dhruvpatel20 (approved):

PR #229, opened 2019-01-16
Bulk print updates

ikenna-hmcts:

- Updated grantCover template

- Updated addAllUserRoles to contain system user

- Update docker-compose config for send-letter service and database

```
[ ] Yes
[x ] No
```


docker/docker-compose-with-ccd.yml
```diff
@@ -3,61 +3,65 @@ networks:
...
     links:
     - send-letter-database
     depends_on:
-    - send-letter-database
+      send-letter-database:
+        condition: service_healthy
 
   send-letter-database:
-    image: docker.artifactory.reform.hmcts.net/reform/send-letter-database
+    image: postgres:9.6-alpine
```
reddy1praveen: try to use same `shared-database:` if possible.
I think you can create table in it.

reddy1praveen (changes requested):

docker/docker-compose-with-ccd.yml
ikenna-hmcts: If you mean the init0-db.sh then no. Could not find a way to make it work as the bulk print db needs specific configs ( ROLE, PASSWORD, DATABASE) are all different.

ikenna-hmcts: Will try to create the database from there.

ikenna-hmcts: Also can't create table from the script as the db needs to exist

malinovsky: I agree with Praveen, we should reuse `shared-database` container. It already has 4 CCD and 1 IDAM database, I can't see why we can't have `send-letter` database there.

ikenna-hmcts: It's now using the shared-database. Thanks

malinovsky (approved):

PR #238, opened 2019-01-23
Bulk print updates

ikenna-hmcts:

### Change description ###
Added Extra logs
Updated docker-compose

```
[ ] Yes
[ x] No
```


src/main/resources/application.yml
```diff
@@ -197,7 +197,7 @@ document:
 
 idam:
   s2s-auth:
-    url: ""
+    url: ${PROVIDER_SERVICE_CLIENT_URL:http://localhost:4502}
```
reddy1praveen, line 200: Sorry to say this we are having too many s2s client.
why can't we use only one.
idam:
  s2s-auth:
    url: ${PROVIDER_SERVICE_CLIENT_URL:http://localhost:4502}

idam.service.host: ${PROVIDER_SERVICE_CLIENT_URL:http://localhost:4502}
auth.provider.service.client:
  key: AAAAAAAAAAAAAAAA
  baseUrl: http://localhost:4502

reddy1praveen (changes requested):

src/main/resources/application.yml
reddy1praveen: i think we can remove this
idam:
  s2s-auth:
    url: ""
    url: ${PROVIDER_SERVICE_CLIENT_URL:http://localhost:4502} i am seeing it used in code.

reddy1praveen (changes requested):

src/main/resources/application.yml
RebeccaBaker: This will be sorted, unfortunately a library we now use service-auth-provider-client requires 

idam:
s2s-auth:
url: ${PROVIDER_SERVICE_CLIENT_URL:http://localhost:4502}

so we will have to refactor our code base to use just the one referenced above

auth.provider.service.client:
key: AAAAAAAAAAAAAAAA
baseUrl: http://localhost:4502
is also required by another library too 

Due to bulk print and investigating issues, I will push this code, and refactor in another PR

RebeccaBaker (approved):

PR #239, opened 2019-01-23
PRO-4570: Update to fix generating a grant preview.

davejones74:
https://tools.hmcts.net/jira/browse/PRO-4570

### Change description ###
Simple fix to add missing caseType to getResponseCaseData method as completing a case was resulting in the caseType value being lost. This had the effect of producing a NullPointerException during generating a grant document. 

Previous cases with the missing caseType can be restored by moving to another state by the caseworker, although this does seem to default to the grant of probate 'gop' - not entirely sure if this is accurate but as these others are new case types, I think it should be fine.

```
[ ] Yes
[x] No
```


sanjayparekh-hmcts (approved):

davejones74:
PRO-4570: Merging fix to cnp

PR #242, opened 2019-01-25
PRO-4297 bulk print

RebeccaBaker:

### Change description ###
Added further logging statements
Retrieve document using binary url and providing user-id in the request header
Updated logic in coversheet

```
[ ] Yes
[X] No
```


sanjayparekh-hmcts (approved):
Be mindful that we've left in those encoded doc dumps - OK for now as dbug

PR #250, opened 2019-01-31
PRO-3936: Update to pass through signature to grant documents.

davejones74:
https://tools.hmcts.net/jira/browse/PRO-3936

### Change description ###
New signature appears on issued grants
Intestacy - Letter of administration
Digital Grant - Grant of Probate
Admon Will - Letter of administration with will
Will Lodgement Receipt Deposit

```
[ ] Yes
[x] No
```


davejones74:
Closing PR to add the missing encryption routines for signature image (which I thought was part of another ticket). Will reopen PR once this is once.

src/main/resources/application.yml
```diff
@@ -79,7 +80,7 @@ spring.jackson.serialization.write_dates_as_timestamps: false
 
 ####Evidence management Service URL's
 #hard coded to point to docker container ccd-data-store-api CCD_DM_DOMAIN value
-evidence.management.host: http://localhost:8080
+evidence.management.host: http://dm-store:8080
```
RebeccaBaker: Please revert back to localhost not dm-store

RebeccaBaker (approved):

src/main/java/uk/gov/hmcts/probate/service/template/pdf/PDFManagementService.java
```diff
@@ -39,11 +51,15 @@
     private final ObjectMapper objectMapper;
     private final HttpServletRequest httpServletRequest;
     private final PDFServiceConfiguration pdfServiceConfiguration;
+    private final FileSystemResourceService fileSystemResourceService;
+    
+    static final String SIGNATURE_DECRYPTION_IV = "P3oba73En3yp7ion";
```
sanjayparekh-hmcts, line 56: we may want to store this on vault too

davejones74: The Initialisation Vector doesn't need to be a secret value (I.e. stored in the Vault). From what I've read it's best to change the IV for each encryption which I'm documenting in confluence but that relies on updating the encrypted image each time with the new IV which isn't that practical in our use case.

From Wikipedia...
An initialization vector has different security requirements than a key, so the IV usually does not need to be secret.

PR #260, opened 2019-02-14
PRO-4670-Created elastic search query for cases with grants issued

kurt-uk:
https://tools.hmcts.net/jira/browse/PRO-4670

### Change description ###
Created ES search query searching for cases that have a probate document of digitalGrant and a issue date of yesterday

```
[ ] Yes
[x] No
```


malinovsky (commented):
@kurt-uk As I already mentioned, you shouldn't use case matching specific `uk.gov.hmcts.probate.model.ccd.raw.casematching.CaseData` class for your requirements. Please use `uk.gov.hmcts.probate.model.ccd.raw.request.CaseData` instead. It would be nice if you create new service for your code as it's not related to case matching.

PR #282, opened 2019-02-25
PRO-4820 BO - case-type not populating in case-matching results

malinovsky:


src/main/java/uk/gov/hmcts/probate/model/CaseType.java
```diff
@@ -29,4 +30,14 @@ public String getName() {
     public static List<CaseType> getAll() {
         return Arrays.asList(CaseType.values());
     }
+
+    public static Optional<CaseType> fromCode(String code) {
```
sanjayparekh-hmcts, line 34: Does this need to be optional? if the case type cannot be found isnt that an exception?

PR #338, opened 2019-03-18
PRO-4941 Legacy Import Alias

RebeccaBaker:
https://tools.hmcts.net/jira/browse/PRO-4941

### Change description ###
Updated to use pipe character for alias separator character as opposed to a comma
Increased test coverage 
Updated imported legacy data for local 

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/probate/service/probateman/mapper/AliasNameMapper.java
```diff
@@ -7,6 +7,7 @@
 import uk.gov.hmcts.reform.probate.model.cases.CollectionMember;
```
sanjayparekh-hmcts: It doesn't look like we are using this mapper at all
If so, we should remove it and its Unit Test

RebeccaBaker: done

sanjayparekh-hmcts (approved):

PR #397, opened 2019-04-09
refactored http client to feign client

kurt-uk:

https://tools.hmcts.net/jira/browse/PRO-4669

### Change description ###
refactored http client to feign client, and removed the dependency for the old httpclient from tests that needed the config

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/probate/config/ApplicationConfiguration.java
```diff
@@ -60,4 +64,19 @@ public Parser parser() {
     public HtmlRenderer htmlRenderer() {
         return HtmlRenderer.builder().build();
     }
+
+    private CloseableHttpClient getHttpClient() {
+        int timeout = 10000;
```
sanjayparekh-hmcts: can we put this timeout value in app.yml?

src/main/java/uk/gov/hmcts/probate/service/UserService.java
```diff
@@ -0,0 +1,65 @@
+package uk.gov.hmcts.probate.service;
```
sanjayparekh-hmcts: maybe we should rename this class to something like
IdamAuthenticateUserService?

src/main/java/uk/gov/hmcts/probate/config/ApplicationConfiguration.java
```diff
@@ -36,7 +38,9 @@ public ResourceBundleMessageSource resourceMessageSource() {
 
     @Bean
     public RestTemplate restTemplate() {
-        return new RestTemplate();
+        RestTemplate restTemplate = new RestTemplate();
+        restTemplate.setRequestFactory(new HttpComponentsClientHttpRequestFactory(getHttpClient()));
```
sanjayparekh-hmcts: I believe Feign uses the restTemplate declared here - thats why you have needed to add the factory in there. Did all your regressions work OK?

sanjayparekh-hmcts (changes requested):

src/main/java/uk/gov/hmcts/probate/config/ApplicationConfiguration.java
```diff
@@ -47,7 +53,7 @@ public ObjectMapper objectMapper(ObjectMapper objectMapper) {
         objectMapper.registerModule(module);
```
dhruvpatel20: instead of modifying the existing objectmapper, could you create a config for feign like this:
https://github.com/hmcts/probate-orchestrator-service/blob/master/src/main/java/uk/gov/hmcts/probate/client/SubmitServiceConfiguration.java
You can create you new object mapper with registered modules in there.

src/main/java/uk/gov/hmcts/probate/config/ApplicationConfiguration.java
dhruvpatel20: I think it would be better if you can set this configuration directly on feign without manipulating rest template - it will require some research

PR #420, opened 2019-04-25
added feign client to push files to azure ftp

kurt-uk:

https://tools.hmcts.net/jira/browse/PRO-4675

### Change description ###
added feign client to push files to azure ftp, added signature to yml that has to be replaced locally to work. modified solution to add grants for day to file rather than a grant per file. adjusted excela and ironMountian to reflect this. Added scheduler annotation to notification config that was removed by a different commit.

```
[ ] Yes
[x] No
```


src/main/resources/application.yml
```diff
@@ -3,7 +3,12 @@ server:
 
 resttemplate:
   httpclient:
-    timeout:10000
+    timeout: 10000
+
+ftp.client:
+  url: https://green.blob.core.windows.net
```
sanjayparekh-hmcts, line 9: does this need to be in vault too?

src/main/java/uk/gov/hmcts/probate/service/FileTransferService.java
```diff
@@ -0,0 +1,59 @@
...
+import java.nio.file.Files;
+
+@Slf4j
+public class FileTransferService {
+
+    private final FileTransferApi fileTransferApi;
+    private static final String VALID_FROM = "2018-03-28";
+    private static final String SS = "b";
+    private static final String SRT = "sco";
+    private static final String SP = "rwdlac";
```
sanjayparekh-hmcts, line 20: do we know what these codes are for? perhaps we can use a more descriptive name for them

src/main/java/uk/gov/hmcts/probate/service/NotificationService.java
```diff
@@ -144,13 +145,13 @@ private Document getGeneratedSentEmailDocument(SendEmailResponse response, Strin
         return personalisation;
     }
 
-    private Map<String, String> getExcelaPersonalisation(Long id, CaseData caseData) {
+    private Map<String, String> getExcelaPersonalisation(List<ReturnedCaseDetails> cases) {
         HashMap<String, String> personalisation = new HashMap<>();
 
+        StringBuilder data = getBuiltData(cases);
+
         personalisation.put("excelaName", LocalDateTime.now().format(EXCELA_DATE) + "will");
```
sanjayparekh-hmcts: probably want these personalisation keys as final const String too

sanjayparekh-hmcts (changes requested):

src/main/resources/application.yml
kurt-uk: apparently not, i asked them what we would need to keep private, no one can do anything as long as they dont have the signature, 
however if you feel i should put them there then i can do that too!

sanjayparekh-hmcts: Nope!

src/main/java/uk/gov/hmcts/probate/service/FileTransferService.java
kurt-uk: not a single clue, i asked James what they were for and it seems that they dont know either, i apparently just need them, which i do. however i tried to include some descriptive names from what i interpreted such as the valid from date which should have been `sv`. unfortunately they couldnt inform me of their purpose so i dont know what to call them other than their paramter value

src/main/java/uk/gov/hmcts/probate/service/NotificationService.java
kurt-uk: sure thing

RebeccaBaker (approved):

PR #520, opened 2019-06-14
PRO-5621: added empty complex type tabs for xls generation

sanjayparekh-hmcts:


sanjayparekh-hmcts:
This change has already been added to master

PR #553, opened 2019-07-10
PRO-5663: Updated to show error if using attach scan docs from UI.

davejones74:
https://tools.hmcts.net/jira/browse/PRO-5663

### Change description ###
Removes attachScannedDocs from CaseEventToFields as it is not necessary for the user to manage scanned documents through the UI. Updated back-office code to add a new endpoint /documents/attach-scanned-docs-error which displayed an error message if the user attempts to select Attach scanned docs' as a Next step.

```
[ ] Yes
[x] No
```


src/test/java/uk/gov/hmcts/probate/controller/DocumentControllerTest.java
```diff
@@ -345,4 +349,17 @@ public void generateGrantDraftReissueGrantOfRepresentation() throws Exception {
                 .andExpect(content().string(containsString("data")));
 
     }
+
+    @Test
+    public void displayAttachScanDocErrorIfUsedFromUI() throws Exception {
+
+        String solicitorPayload = testUtils.getStringFromFile("solicitorPayloadNotifications.json");
+
+        ResultActions resultActions = mockMvc.perform(post("/document/attach-scanned-docs-error")
```
kurt-uk: whats with the ResultActions? its not used anywhere and it looks like youre just getting the body of the message, could the methods that are already used in the other tests be used here too??

src/main/java/uk/gov/hmcts/probate/controller/DocumentController.java
```diff
@@ -233,4 +235,18 @@
         return ResponseEntity.ok(callbackResponseTransformer.addDocuments(callbackRequest,
                 Arrays.asList(document), null, null));
     }
+
+    @PostMapping(path = "/attach-scanned-docs-error", consumes = MediaType.APPLICATION_JSON_UTF8_VALUE)
+    public ResponseEntity<CallbackResponse> displayAttachScanDocErrorIfUsedFromUI(@RequestBody CallbackRequest callbackRequest) {
```
kurt-uk: take a look at the default exception handler class

src/test/java/uk/gov/hmcts/probate/controller/DocumentControllerTest.java
davejones74: I've removed ResultActions, as you said it wasn't used.

src/main/java/uk/gov/hmcts/probate/controller/DocumentController.java
davejones74: Updated with Sanjay's example in CaveatQueryService. Also changed the DefaultExceptionHandler to correct the order or messages.

src/main/java/uk/gov/hmcts/probate/exception/handler/DefaultExceptionHandler.java
```diff
@@ -58,7 +58,7 @@
     public ResponseEntity<CallbackResponse> handle(BusinessValidationException exception) {
         log.warn(exception.getMessage());
         CallbackResponse callbackResponse = CallbackResponse.builder()
-                .errors(Collections.singletonList(exception.getMessage()))
+                .errors(Collections.singletonList(exception.getUserMessage()))
```
kurt-uk, line 61: 👍

src/main/java/uk/gov/hmcts/probate/controller/ErrorController.java
```diff
@@ -0,0 +1,37 @@
...
+import static org.springframework.http.MediaType.APPLICATION_JSON_UTF8_VALUE;
+import static org.springframework.http.MediaType.APPLICATION_JSON_VALUE;
+
+@Slf4j
+@RequiredArgsConstructor
+@RequestMapping(value = "/error", consumes = APPLICATION_JSON_UTF8_VALUE, produces = APPLICATION_JSON_VALUE)
+@RestController
+public class ErrorController {
+
+    private final BusinessValidationMessageRetriever businessValidationMessageRetriever;
```
RebeccaBaker: ErrorController is very generic, can we make it more specific like BulkScanningController as it is related to a bulk scanning event. Then within this class we have the method error

src/main/java/uk/gov/hmcts/probate/controller/BulkScanningController.java
```diff
@@ -0,0 +1,37 @@
...
+import uk.gov.hmcts.probate.service.BusinessValidationMessageRetriever;
+
+import java.util.Locale;
+
+import static org.springframework.http.MediaType.APPLICATION_JSON_UTF8_VALUE;
+import static org.springframework.http.MediaType.APPLICATION_JSON_VALUE;
+
+@Slf4j
+@RequiredArgsConstructor
+@RequestMapping(value = "/error", consumes = APPLICATION_JSON_UTF8_VALUE, produces = APPLICATION_JSON_VALUE)
```
kurt-uk: this path and the path on line 29 should be switched around, the class base is not error anymore and is bulk scanning/scanned docs so the base should reflect that with the method of the error handling then using the path of `error`

src/main/java/uk/gov/hmcts/probate/controller/BulkScanningController.java
```diff
@@ -0,0 +1,37 @@
...
+import uk.gov.hmcts.probate.service.BusinessValidationMessageRetriever;
+
+import java.util.Locale;
+
+import static org.springframework.http.MediaType.APPLICATION_JSON_UTF8_VALUE;
+import static org.springframework.http.MediaType.APPLICATION_JSON_VALUE;
+
+@Slf4j
+@RequiredArgsConstructor
+@RequestMapping(value = "/bulk-scan", consumes = APPLICATION_JSON_UTF8_VALUE, produces = APPLICATION_JSON_VALUE)
```
kurt-uk:
```suggestion
@RequestMapping(value = "/bulk-scanning", consumes = APPLICATION_JSON_UTF8_VALUE, produces = APPLICATION_JSON_VALUE)
```

RebeccaBaker (approved):

PR #596, opened 2019-07-31
PRO-5663: Revert attach scanned doc callback error message

davejones74:
https://tools.hmcts.net/jira/browse/PRO-5663

### Change description ###
Adding a callback for attachScannedDocs to provide an error message to the user if using this event from the UI breaks the use of the event from the Bulk Scanning Orchestrator. This change corrects the demo environment to remove the callback from CCD (just installed in demo at the moment) and remove the callback endpoint in the back office code.

The previous update also removed the CaseEventToFields for attachScannedDocs so at least the user will just get an empty 'continue' page if they chose to select 'Attach a Scanned Doc' by accident.

```
[ ] Yes
[x] No
```


kurt-uk (commented):
i thought you said you wasnt deleting BS controller and controller test?

kurt-uk (approved):

PR #679, opened 2019-09-10
Pro 6127 update bsp ocr mandatory keys

davejones74:
https://tools.hmcts.net/jira/browse/PRO-6127

### Change description ###
Update mandatory fields after review with product owner.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/probate/service/ocr/OCRToCCDMandatoryField.java
```diff
@@ -1,47 +1,118 @@
...
+                        descriptions.add(String.format(MANDATORY_FIELD_WARNING_STIRNG, field.getKey()));
                     }
                 });
                 break;
             case PA1A:
+                Stream.of(IntestacyMandatoryFields.values()).forEach(field -> {
+                    log.info("Checking {} against ocr fields", field.getKey());
+                    if (!ocrFieldValues.containsKey(field.getKey())) {
+                        log.warn("{} was not found in ocr fields", field.getKey());
+                        descriptions.add(String.format(MANDATORY_FIELD_WARNING_STIRNG, field.getKey()));
```
kurt-uk: no longer needs to be description if we are not having descriptions anymore

davejones74: The descriptions List used here is to build the list of warnings and is necessary. The description property on the OCRField isn't necessary, but could be kept so that it can be used once the description field is supported by bulk scanning. Let me know if you would like me to remove this property (from OCRField) and I can update the code.

kurt-uk (approved):

PR #751, opened 2019-10-09
Pro 5804 validate and transform er

davejones74:
https://tools.hmcts.net/jira/browse/PRO-5804

### Change description ###
Adds a new endpoint to probate-back-office to transform Caveat OCR data from an exception record to allow case workers to create cases from scanned form data.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/probate/controller/ExceptionRecordController.java
```diff
@@ -0,0 +1,115 @@
...
+                    .errors(errors)
+                    .build();
+
+        } else {
+            switch (formType) {
+                case PA8A:
+                    callbackResponse = erService.createCaveatCaseFromExceptionRecord(erRequest, warnings);
+                    return ResponseEntity.ok(callbackResponse);
+                default:
+                    // Do nothing
```
RebeccaBaker: do we not need break; or return statement here?

RebeccaBaker (approved):

PR #799, opened 2019-11-19
PRO-5894 send copy of will

RebeccaBaker:
https://tools.hmcts.net/jira/browse/PRO-6484
https://tools.hmcts.net/jira/browse/PRO-5984

### Change description ###
PRO-6484 - update bulk print call to allow more than 30 copies 
PRO-5984 - retrieve scanned will and apply watermark, save to case data, then send sealed will to bulk print

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/probate/controller/DocumentController.java
```diff
@@ -169,15 +140,24 @@
             documentService.expire(callbackRequest, documentType);
         }
 
+        List<Document> documents = new ArrayList<>();
```
sanjayparekh-hmcts, line 144: I would like to minimise the logic in the controllers if possible. Maybe split out the specific doc generation paths into some other services?

src/main/java/uk/gov/hmcts/probate/model/Constants.java
```diff
@@ -14,11 +14,14 @@
     public static final String EXTRACT_ADMON_WILL = "ADMON/WILL";
```
sanjayparekh-hmcts, line 14: DO these ultimately belong somewhere in commons?

PR #876, opened 2020-01-16
Pro 6868 clear grant specific data when changing grant type

jackmunkonge:
https://tools.hmcts.net/jira/browse/PRO-6868

### Change description ###
clear grant specific data when changing grant type

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/probate/transformer/CallbackResponseTransformer.java
```diff
@@ -85,6 +88,10 @@
     private static final String READY_FOR_EXAMINATION = "BOReadyForExamination";
     private static final String EXAMINING = "BOExamining";
     private static final String NO_WILL = "NoWill";
+    private static final String WILL_LEFT = "WillLeft";
```
sanjayparekh-hmcts: are any of these constants being setup already somewhere else?

src/main/java/uk/gov/hmcts/probate/transformer/CallbackResponseTransformer.java
```diff
@@ -1082,6 +1096,56 @@ private String getSolsSOTName(String firstNames, String surname) {
         return sb.toString();
     }
 
+    private ResponseCaseDataBuilder clearGrantSpecificData(CallbackRequest callbackRequest, ResponseCaseDataBuilder responseCaseDataBuilder, String newState) {
```
sanjayparekh-hmcts: might be a good idea to move this into its own service, to lessen the burden on the transformer - not essential

sanjayparekh-hmcts (changes requested):

src/main/java/uk/gov/hmcts/probate/transformer/TransformerUtils.java
```diff
@@ -0,0 +1,157 @@
...
+
+    public static String getSolsSOTName(String firstNames, String surname) {
+        StringBuilder sb = new StringBuilder();
+        sb.append(firstNames);
+        sb.append(" " + surname);
+        return sb.toString();
+    }
+
+
+    public static ResponseCaseData.ResponseCaseDataBuilder clearGrantSpecificData(CallbackRequest callbackRequest, ResponseCaseData.ResponseCaseDataBuilder responseCaseDataBuilder, String newState) {
```
sanjayparekh-hmcts: I was actually thinking to just put this method in its own service. I dont really like the idea of having a class called... Utils, better to be one specific function - but leave that for now + just create this method in its own service, referenced from the CallbackTransformer

sanjayparekh-hmcts (approved):

PR #1039, opened 2020-04-05
PRO-1234: corrected hmrc de to 1M case max

sanjayparekh-hmcts:
update Data Extract Query to retrieve 1M cases maximum


jackmunkonge (approved):

ruban72 (commented):
Reviewed changes

PR #1156, opened 2020-09-23
Update dev env

linusnorton:
Update the compose file to be compatible with the npx tool.

One small backwards compatible change to detect when the functional tests are executing against a docker env and correct the expected url back from dm-store.


src/functionalTest/java/uk/gov/hmcts/probate/functional/util/FunctionalTestUtils.java
```diff
@@ -97,7 +100,7 @@ public String downloadPdfAndParseToString(String documentUrl) {
         Response document = RestAssured.given()
                 .relaxedHTTPSValidation()
                 .headers(getHeadersWithUserId())
-                .when().get(documentUrl).andReturn();
+                .when().get(documentUrl.replace("http://dm-store:8080", dmStoreUrl)).andReturn();
```
sanjayparekh-hmcts, line 103: shouldnt we just be making sure the right value is passed in here for the relevant env?

linusnorton: when it runs inside a docker container the dm-store (and CCD) use the docker domain name (dm-store), the functional tests execute outside the docker environment and need to download that file so it has to translate the doc url.

sanjayparekh-hmcts (approved):

PR #1159, opened 2020-09-24
Dtspb 567 wrap null exceptions

SarahLittlejohn:
https://tools.hmcts.net/jira/browse/DTSPB-567

### Change description ###
Throw exception for null value.

```
[ ] Yes
[ X ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/probate/exception/ClientDataException.java
+++ b/src/main/java/uk/gov/hmcts/probate/exception/ClientDataException.java
@@ -2,37 +2,12 @@ package uk.gov.hmcts.probate.exception;
 
 import lombok.Data;
 import lombok.EqualsAndHashCode;
-import org.springframework.validation.Errors;
-import uk.gov.hmcts.probate.exception.model.FieldErrorResponse;
-
-import java.util.List;
-import java.util.Optional;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 @Data
 @EqualsAndHashCode(callSuper = true)
 public class ClientDataException extends RuntimeException {
 
-    private final List<FieldErrorResponse> errors;
-
-    public ClientDataException(final String message, final Errors errors) {
-        super(message);
-        this.errors = Optional.ofNullable(errors)
-                .map(Errors::getFieldErrors)
-                .map(List::stream)
-                .orElse(Stream.empty())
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/hmcts/probate/service/CaveatQueryService.java
```diff
@@ -81,8 +84,7 @@ public CaveatData findCaveatById(CaseType caseType, String caveatId) {
         }
         return foundCaveats.get(0).getData();
     }
-
-    @Nullable
+
```
isha2040, line 75: can we please get ride of this new line and keep same as master

isha2040 (approved):

src/main/java/uk/gov/hmcts/probate/service/CaveatQueryService.java
```diff
@@ -106,14 +96,21 @@ public CaveatData findCaveatById(CaseType caseType, String caveatId) {
...
+            throw new ClientDataException(e.getMessage());
         }
 
-        appInsights.trackEvent(REQUEST_SENT, uri.toString());
 
+        appInsights.trackEvent(REQUEST_SENT, uri.toString());
         return returnedCaveats.getCaveats();
     }
+
+    private static <T> T nonNull(@Nullable T result) {
```
sanjayparekh-hmcts, line 112: do we really need @Nullable here?

src/main/java/uk/gov/hmcts/probate/exception/ClientDataException.java
```diff
@@ -0,0 +1,38 @@
...
+    public ClientDataException(final String message, final List<FieldErrorResponse> errors) {
+        super(message);
+        this.errors = errors;
+    }
+
+    public ClientDataException(final String message) {
+        super(message);
+        this.errors = null;
+    }
+}
```
sanjayparekh-hmcts, line 13: do you need a unit test on this class too?

src/main/java/uk/gov/hmcts/probate/service/CaveatQueryService.java
isha2040: its just to indicate that the value passed could have null values , if you suggest to get rid of it can try that out as well

sanjayparekh-hmcts (approved):

PR #1175, opened 2020-10-08
strip \r from text before comparing in tests

vinceholt:
### Change description ###
Strip \r within testcode for both expected and actual output, in order for tests to run successfully on Windows as well as Linux/Mac os.

```
[ ] Yes
[ x] No
```


vinceholt:
Was actually down to local git config automatically converting LFs (\n) to CRLFs (\r\n). Fixed by attending local config.

PR #1189, opened 2020-10-21
DTSPB-633: corrected xls and code to accept selected paper form value

sanjayparekh-hmcts:
corrected xls and code to accept selected paper form value


src/main/java/uk/gov/hmcts/probate/transformer/CallbackResponseTransformer.java
```diff
@@ -469,7 +469,6 @@ private boolean doTransform(CallbackRequest callbackRequest) {
     public CallbackResponse paperForm(CallbackRequest callbackRequest) {
 
         ResponseCaseDataBuilder responseCaseDataBuilder = getResponseCaseData(callbackRequest.getCaseDetails(), false);
-        responseCaseDataBuilder.paperForm(ANSWER_YES);
```
KleoG, line 472: Is paper form mandatory? doesn't it need to be assigned a value?

sanjayparekh-hmcts: i removed that line because we NOW want to take the value chosen by the user

KleoG (approved):

PR #1216, opened 2020-11-17
DTSPB-778 Case Progress Tab

vinceholt:
### JIRA links ###
https://tools.hmcts.net/jira/browse/DTSPB-715
https://tools.hmcts.net/jira/browse/DTSPB-776
https://tools.hmcts.net/jira/browse/DTSPB-777
https://tools.hmcts.net/jira/browse/DTSPB-778
https://tools.hmcts.net/jira/browse/DTSPB-779
https://tools.hmcts.net/jira/browse/DTSPB-780
https://tools.hmcts.net/jira/browse/DTSPB-781
https://tools.hmcts.net/jira/browse/DTSPB-784

### Change description ###
Add Case Progress Tab for Professional Users. to show custom html for:
1) Stopped State
2) Escalated to Registrar State
3) For any other state, a Task List, showing which sections of the application are complete, incomplete or not started

```
[ ] Yes
[ x] No
```


.vscode/launch.json
```diff
@@ -0,0 +1,68 @@
+{
```
meganaslattery, line 1: Not sure we want a vscode file added to the code base?

meganaslattery (approved):
It will have to go in after DTSPB-922 but it looks good!

.vscode/launch.json
vinceholt: ah yes went through this with Jack, the jury is out on this on the internet, with camps for and against; but the aginst camp is the one where lots of individuals have their own launch.json entries for their own purposes, we don't have that, and mor often than not if you want to run something from package.json in the debugger, you'd want others to be able to do that too - e.g. mocha test an individual file. So I think in our context, just invoking package.json entries, it's beneficial for others to be able to inherit shared invocations to save having to manually type in a terminal window, and we're unlikely to have entries other than package.json invoke. Doesn't harm or affect user of other IDEs

vinceholt:
replaced with a new pr for the epic

PR #1230, opened 2020-12-03
DTSPB-988: reinstated solsFeeAccountNumber on copies tab

sanjayparekh-hmcts:
reinstate solsFeeAccountNumber on copies tab
(also renamed copies tabe id to lowe case copies)


elvianihuseyin (approved):
LGTM

PR #1362, opened 2021-02-22
Dtspb 910 add trust corps other execs page

meganaslattery:

https://tools.hmcts.net/jira/browse/DTSPB-910

### Change description ###

About the executor pages trust corps

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/probate/model/ccd/raw/response/ResponseCaseDataParent.java
```diff
@@ -17,44 +17,48 @@
 @Data
 public class ResponseCaseDataParent {
 
-    protected String schemaVersion;
+    protected final String schemaVersion;
     // A second copy of schemaVersion, holding exactly the same value.
```
vinceholt, line 21: Does it work if this is set as a constant? Is set when a rest api is invoked using builder - if it works ok then schemaVersionCcdCopy should be final, and final string declarations (constants) should be grouped together

vinceholt (commented):
Just the one comment re: final, but also - don't end to end tests need to be updated?

KleoG (approved):

vinceholt (approved):

src/main/java/uk/gov/hmcts/probate/model/ccd/raw/response/ResponseCaseDataParent.java
meganaslattery: This should work fine!

PR #1496, opened 2021-04-20
Match journey with Miro for 673

stelioskalo:
### Change description ###
Match the journey exactly with the Miro board

```
[ ] Yes
[X] No
```


vinceholt (commented):
Just a query on lodgement address in e2e tests - think we should still see this?

vinceholt (approved):

PR #1605, opened 2021-06-15
DTSPB-2098 null parent options

stelioskalo:
https://tools.hmcts.net/jira/browse/DTSPB-2098

### Change description ###
Nullify parent fields so as to hide executor lists

```
[ ] Yes
[X] No
```


vinceholt:
This has now been withdrawn after discussion with Sanjay

PR #1815, opened 2022-02-03
Dtspb 2006 senddocs del

DelythJustice:
https://tools.hmcts.net/jira/browse/DTSPB-2006

DTSPB-2006 sendDocs non English will changes for including authorised translation of the original will - first pass

```
[ ] Yes
[ x] No
```


docker-compose.yml
```diff
@@ -98,21 +98,6 @@ services:
       - sidam-api
       - service-auth-provider-api
 
-  ccd-case-management-web:
```
AbdullahShahHMCTS: will need to roll back this file

DelythJustice: done, thanks

AbdullahShahHMCTS (approved):
LGTM

src/functionalTest/java/uk/gov/hmcts/probate/functional/documents/SolBaCcdServiceDocumentsTests.java
```diff
@@ -2150,6 +2151,14 @@ public void verifyGenerateSolsCoverSheetAdmonWill() {
 
     }
 
+    @Test
+    public void verifyGenerateSolsCoverSheetAuthenticatedTranslationOfWill() {
+        String payload = "/caseprogress/04f-caseCreated.json";
+        String response = getDocumentTextAtPath(payload, VALIDATE_PROBATE_URL, "solsCoversheetDocument");
+        assertTrue(response.contains(AUTHENTICATED_TRANSLATION_WILL_TEXT));
```
sanjayparekh-hmcts: I think I would prefer this AUTHENTICATED_TRANSLATION_WILL_TEXT value to be hard-coded in the test itself, because if the value changes for some reason, we want the test to highlight it... same for others below too

src/main/java/uk/gov/hmcts/probate/service/template/markdown/MarkdownDecoratorService.java
```diff
@@ -32,4 +36,11 @@ public String getPA17FormLabel(CaseData caseData) {
         }
         return "";
     }
+
+    public String getAuthenticatedTranslationLabel(CaseData caseData) {
+        if (authenticatedTranslationBusinessRule.isApplicable(caseData)) {
+            return "\n*" + AUTHENTICATED_TRANSLATION_WILL_TEXT;
```
sanjayparekh-hmcts: not sure if it makes a diff to markdown, but the other labels have some spaces after the star?

src/test/java/uk/gov/hmcts/probate/service/template/pdf/SolicitorCoversheetPDFDecoratorTest.java
```diff
@@ -41,6 +45,7 @@ public void setup() {
     
     @Test
     public void shouldNotProvideAdditionalDecoration() {
+        when(authenticatedTranslationBusinessRuleMock.isApplicable(caseDataMock)).thenReturn(false);
```
sanjayparekh-hmcts: Perhaps we should set ALL business rules to return false in the setup method, then only set the true returns in the specific tests?

sanjayparekh-hmcts (changes requested):

src/main/java/uk/gov/hmcts/probate/service/template/markdown/MarkdownDecoratorService.java
DelythJustice: It looks the same on test

DelythJustice: I can put in spaces though

src/test/java/uk/gov/hmcts/probate/service/template/pdf/SolicitorCoversheetPDFDecoratorTest.java
DelythJustice: I think with mockito once the when is set then subsequent calls would return the last when. So we would need to change the Before to a BeforeEach I think?

src/functionalTest/java/uk/gov/hmcts/probate/functional/documents/SolBaCcdServiceDocumentsTests.java
DelythJustice: updated thanks

src/test/java/uk/gov/hmcts/probate/service/template/pdf/SolicitorCoversheetPDFDecoratorTest.java
DelythJustice: updated thanks

src/main/java/uk/gov/hmcts/probate/service/template/markdown/MarkdownDecoratorService.java
DelythJustice: updated thanks

sanjayparekh-hmcts (approved):

PR #1820, opened 2022-02-09
DTSPB-2581: clean up FTs

sanjayparekh-hmcts:


src/functionalTest/java/uk/gov/hmcts/probate/functional/fee/SolCcdServiceFeeTests.java
```diff
@@ -1,60 +1,70 @@
...
     @Test
-    public void verifyNetValue5000() {
-        validatePostRequestSuccessForFee("success.feeNetValue5000.json", "applicationFee", "0");
+    public void shouldValidatePaymentInsufficientFunds() {
+        validatePostRequestFailureForPBAs(1000, "solicitorPDFPayloadProbateCopiesForInsufficientFunds.json",
+            "have insufficient funds available");
     }
 
-    // currently failing after PBA merge - to be reinstated
+    @Pending
```
srochani, line 104: Are we planning to fix this test or just ignoring it

srochani (approved):

src/main/resources/application.yml
```diff
@@ -62,7 +62,7 @@ pba:
 docmosis:
   service:
     uri: ${PDF_SERVICE_BASEURL:https://docmosis.aat.platform.hmcts.net}
-    accessKey: ${PDF_SERVICE_ACCESS_KEY:ZDYxMTkzZTQtMGY2Mi00NDM1LWIyN2ItNGRkNzdjOTczMjAwOjQ1NTE0ODQ}
+    accessKey: ${PDF_SERVICE_ACCESS_KEY:dyVv8pXwQ03RRyJZQIPX2RWP9LgJJGTU08kc9dA8ATJoA9EZXQEWe7L1Uwe}
```
isha2040, line 65: this is read from the secrets, do we still need this key here ? should be ideally changed to something dummy

sanjayparekh-hmcts: if we want to remove this for local, we should do on another ticket

isha2040 (approved):

PR #1845, opened 2022-03-02
added checks for new IHT tab

AhsanZia-Moodys:

```
[x] Yes
[ ] No
```

### DIFF
--- a/src/test/end-to-end/pages/caseDetails/grantOfProbate/ihtTabConfig.json
+++ b/src/test/end-to-end/pages/caseDetails/grantOfProbate/ihtTabConfig.json
@@ -1,0 +1,19 @@
+{
+  "tabName": "IHT",
+  "fields": [
+    "Did you complete IHT forms to report the estate's values?",
+    "Gross value of the IHT estate",
+    "Net value of the IHT estate",
+    "Net qualifying value of the IHT estate",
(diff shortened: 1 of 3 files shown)


Mitesh1909 (changes requested):
Does this change not have to be on grantOfProbatePath.js too?

AhsanZia-Moodys:
> Does this change not have to be on grantOfProbatePath.js too?

I have discovered that the IHT tab is added to the case once the expected estate page is complete. In the normal journey, the expected estate page doesn't get completed, therefore the IHT tab doesn't show up and is not needed in those files.

Mitesh1909 (commented):
Would it not be better to use a EE application than a IHT207? This way we can check the Gross, Net and Net Qualifying text and values?

AhsanZia-Moodys:
> Would it not be better to use a EE application than a IHT207? This way we can check the Gross, Net and Net Qualifying text and values?

Gross, Net and Net qualifying text has been checked in the "src/test/end-to-end/pages/caseDetails/grantOfProbate/ihtTabConfig.json" files. The IHT207 is selected in the EE Yes journey and checked in the tab to see if IHT207 is visible.

Mitesh1909 (approved):

PR #1859, opened 2022-03-10
DTSPB-2512 share a caveat

DelythJustice:
share caveat, tests to follow

```
[ ] Yes
[x] No
```


ccdImports/configFiles/CCD_Probate_Caveat/CaseEvent.json
```diff
@@ -71,8 +71,8 @@
...
-  {"LiveFrom": "01/01/2017", "CaseTypeID": "Caveat", "ID": "solicitorUpdateCaveat", "Name": "Application details", "Description": "Solicitor updates application details", "DisplayOrder": 73, "PreConditionState(s)": "SolAppCreated", "PostConditionState": "SolAppUpdated", "CallBackURLAboutToSubmitEvent": "http://${CCD_DEF_CASE_SERVICE_BASE_URL}/caveat/solsUpdate", "SecurityClassification": "Public", "ShowSummary": "Y", "EndButtonLabel": "Save and continue"},
+  {"LiveFrom": "01/01/2017", "CaseTypeID": "Caveat", "ID": "solicitorCreateCaveat", "Name": "Raise a caveat (Solicitor)", "Description": "Raise a caveat against grant application (Solicitor)", "DisplayOrder": 72, "PostConditionState": "SolAppCreated", "CallBackURLAboutToSubmitEvent": "http://${CCD_DEF_CASE_SERVICE_BASE_URL}/caveat/solsCreate", "CallBackURLSubmittedEvent": "http://${CCD_DEF_CASE_SERVICE_BASE_URL}/case/sols-access","SecurityClassification": "Public", "ShowSummary": "Y", "EndButtonLabel": "Save and continue"},
```
itoulou, line 74: I think the after submit callback is pointing to /case/sols-access which posts a request with CaseDetails on it but I think we want CaveatDetails. So I think you need to create a new endpoint in CaveatController so it hits something like /caveat/sols-access which will post Caveat Details in the request

itoulou: Also is there a need to assign access on solicitorUpdateCaveat event as well? I think a caveat is created after the first event (solicitorCreateCaveat) so no need to assign roles on both events.

src/main/java/uk/gov/hmcts/probate/transformer/CaveatCallbackResponseTransformer.java
```diff
@@ -250,7 +271,8 @@ private ResponseCaveatDataBuilder getResponseCaveatData(CaveatDetails caveatDeta
             .autoClosedExpiry(caveatData.getAutoClosedExpiry())
             .pcqId(caveatData.getPcqId())
             .bulkScanEnvelopes(caveatData.getBulkScanEnvelopes())
-            .payments(caveatData.getPayments());
+            .payments(caveatData.getPayments())
+            .applicantOrganisationPolicy(caveatData.getApplicantOrganisationPolicy());
```
itoulou, line 277: Not sure if this line is needed, as you're already adding to response caveat data in transformForSolicitor() on line 164

ccdImports/configFiles/CCD_Probate_Caveat/CaseEvent.json
DelythJustice: I was planning to write a separate access for caveat but it doesn't need any different information as it only needs the id. I could refactor so that it clearly only takes in the id as opposed to CaseDetails?

DelythJustice: Or a very simple common parent with just the id on it - that's the only not null for both CaseDetails and CaveatDetails

src/main/java/uk/gov/hmcts/probate/transformer/CaveatCallbackResponseTransformer.java
DelythJustice: Thanks, I'll keep it here and remove from the transformForSolicitor - then it matches the probate code

ccdImports/configFiles/CCD_Probate_Caveat/CaseEvent.json
DelythJustice: I'm going to simplify and just pass the caseId so the code can be common rather than duplicated

src/main/java/uk/gov/hmcts/probate/transformer/CaveatCallbackResponseTransformer.java
DelythJustice: oh, realised I do need both, different cases - so I fill in the org details from the token in one and just keep passing it along with the other. As for probate case code.

PR #1863, opened 2022-03-16
fix case progress e2e tests

shravanmechineni:
https://tools.hmcts.net/jira/browse/DTSPB-2650

```
[ ] Yes
[X] No
```


package.json
```diff
@@ -43,9 +43,9 @@
...
-    "test-e2e-caseProgressStandard": "E2E_TEST_PATH=./paths/caseprogress/caseProgressStandardPath.js E2E_TEST_TIME_TO_WAIT_FOR_TEXT=0 E2E_AUTO_DELAY_ENABLED=false BO_E2E_TEST_ACTION_WAIT_TIME=0 NODE_PATH=. node ./node_modules/codeceptjs/bin/codecept.js run-multiple parallel -c ./src/test/end-to-end/ --steps --reporter mochawesome",
-    "test-e2e-caseProgressReenter": "E2E_TEST_PATH=./paths/caseprogress/caseProgressReenterPath.js E2E_TEST_TIME_TO_WAIT_FOR_TEXT=0 E2E_AUTO_DELAY_ENABLED=false BO_E2E_TEST_ACTION_WAIT_TIME=0 NODE_PATH=. node ./node_modules/codeceptjs/bin/codecept.js run-multiple parallel -c ./src/test/end-to-end/ --steps --reporter mochawesome",
-    "test-e2e-caseProgressAppStopped": "E2E_TEST_PATH=./paths/caseprogress/caseProgressApplicationStoppedPath.js E2E_TEST_TIME_TO_WAIT_FOR_TEXT=0 E2E_AUTO_DELAY_ENABLED=false BO_E2E_TEST_ACTION_WAIT_TIME=0 NODE_PATH=. node ./node_modules/codeceptjs/bin/codecept.js run-multiple parallel -c ./src/test/end-to-end/ --steps --reporter mochawesome",
```
isha2040, line 48: why is -multiple parallel removed ?

shravanmechineni: I just removed it while testing it locally and there is no need for multiple parallel as it's just 1 test only (e.g. E2E_TEST_PATH=./paths/caseprogress/caseProgressStandardPath.js)
However, there is no effect on nightly builds as the tests are configured to run from bin/run-e2e-tests.sh script.

isha2040: ok I will take your lead here . Thanks

isha2040 (approved):

Mitesh1909 (commented):
src/test/end-to-end/pages/utility/navigateToCase.js
How come you have removed most of this code? My understanding was that it was there to cater for the issues we had on successfully executing in the pipeline?

shravanmechineni:
> src/test/end-to-end/pages/utility/navigateToCase.js How come you have removed most of this code? My understanding was that it was there to cater for the issues we had on successfully executing in the pipeline?

It seems like there is no option to search a specific case with id now, so I think it's no longer a need to use that code.

Mitesh1909 (approved):

PR #1876, opened 2022-03-29
DTSPB-2647 Validate Bulk Scan Executors Email Address

MarcusHMCTS:

https://tools.hmcts.net/jira/browse/DTSPB-2647

```
[ ] Yes
[X] No
```


src/functionalTest/resources/json/bulkscan/version2/validation/expectedWarnings/invalidApplyingExecutor0EmailAddress.txt
```diff
@@ -0,0 +1 @@
+Second applicant email address (executorsApplying_0_applyingExecutorEmail) does not appear to be a valid email address
```
isha2040, line 1: Is this text gonna appear in the logs ? Would it always be applicable on Second email address. What if 3rd or 4th executer email is invalid and not the second ?

MarcusHMCTS: This error message won't log, and it show on the XUI and the user can correct the OCR problem

For Second email address, this is only a function test case, add by me, all 2nd , 3rd & 4th executor email address are not mandatory,
if any executors email address is invalid, the relative error message will show

The email regex has already used by the primaryApplicantEmailAddress, solsSolicitorEmail & caveatorEmailAddress at OcrEmailValidator.java with the EmailValidationService.java before

sanjayparekh-hmcts (approved):

PR #1877, opened 2022-03-31
DTSPB-2618: Error message in Request information event for paper appl…

srochani:
https://tools.hmcts.net/jira/browse/DTSPB-2618

### Change description ###

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/probate/controller/DocumentController.java
```diff
@@ -306,8 +304,6 @@ private String getPdfSize(@Valid CaseData caseData) {
     // This only seems to be called once list lists are mapped to exec lists
     @PostMapping(path = "/generate-sot", consumes = MediaType.APPLICATION_JSON_VALUE)
     public ResponseEntity<CallbackResponse> generateStatementOfTruth(@RequestBody CallbackRequest callbackRequest) {
-        redeclarationSoTValidationRule.validate(callbackRequest.getCaseDetails());
```
sanjayparekh-hmcts: Sunil - I dont believe the JIRA is asking for us to remove this rule for the generate-sot process - can you check please?

sanjayparekh-hmcts (changes requested):

src/main/java/uk/gov/hmcts/probate/controller/DocumentController.java
srochani: redeclarationSoTValidationRule is checks for paper case and throw error, as per ticket we have to remove this validation error, after removing the validation error this class would be empty so I have removed the rules itself

sanjayparekh-hmcts (approved):

PR #1970, opened 2022-06-24
Setup CCD Diff

linusnorton:
@DelythJustice I've set up CCD diff. Once merged into master PRs will receive a report on the PR like this one: https://github.com/hmcts/rse-ccd-diff/pull/4#issuecomment-1163538298

Everyone's JSON file structure is a little bit different but I believe it will work with yours. Let me know if there are any issues or you notice that it's not picking up some changes.

### DIFF
--- a/.github/workflows/ccd-diff.yml
+++ b/.github/workflows/ccd-diff.yml
@@ -1,0 +1,67 @@
+name: Diff CCD definition
+
+on:
+  pull_request:
+    branches:
+      - master
+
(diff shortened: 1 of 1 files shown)


DelythJustice:
Thanks very much for this. Might need to wait until Monday to merge, business usually block fridays.

sanjayparekh-hmcts (approved):

DelythJustice (approved):
fab

PR #2035, opened 2022-09-02
DTSPB-2683-2987-Amend-GOP-LegalStatement

FeliTam:
https://tools.hmcts.net/jira/browse/DTSPB-2683
https://tools.hmcts.net/jira/browse/DTSPB-2987

### Change description ###
Amend GOP Legal Statement wordings to react on willNumberOfCodicils
in both CW and Solicitor flow

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/probate/transformer/CallbackResponseTransformer.java
```diff
@@ -1363,7 +1363,8 @@ private void updateCaseBuilder(CaseData caseData, ResponseCaseDataBuilder<?, ?>
 
         if (!isCodicil(caseData)) {
             builder
-                    .willNumberOfCodicils(null);
+                    .willNumberOfCodicils(null)
+                    .codicilAddedDateList(null);
```
sanjayparekh-hmcts: I dont understand, why are we setting this list to null?

src/main/resources/templates/pdf/legalStatementProbate.html
```diff
@@ -44,6 +44,7 @@
 </head>
 <body>
 <table>
+    {% set numberOfCodicils = case_details.case_data.codicilAddedDateList|length %}
```
sanjayparekh-hmcts, line 47: we are probably better off aligning these HTML changes with what Iswarya has done for her JIRAs
I think she takes a null value into account
Please check with her

src/main/java/uk/gov/hmcts/probate/transformer/CallbackResponseTransformer.java
FeliTam: It is to fix a bug I found during testing.
Scenario:
User put "Yes" to "Were any codicils added", add codicil date and then Submit.
User amend detail again and put "No" to "Were any codicils added" without removing codicil date manually and then Submit.
Legal statement generated incorrectly as the codicil date list is not empty.

src/main/resources/templates/pdf/legalStatementProbate.html
FeliTam: Aligned template with changes made by Iswarya

PR #2110, opened 2022-12-15
DTSPB-3237 stop trying to create roles that exist in idam already

DelythJustice:
DTSPB-3237

### Change description ###
stop trying to create roles that exist in idam already and locking AAT account as we use wrong password

```
[ ] Yes
[x] No
```


FeliTam (commented):
It is the same as add-roles.sh. Should it be removing all lines with executing idam-role.sh?

FeliTam (approved):

PR #2118, opened 2023-01-10
DTSPB-3264 - Optimise cftlib local setup

danielbe11:

https://tools.hmcts.net/jira/browse/DTSPB-3264

### Change description ###

- Run BO, XUI and CCD services using bootWithCcd. Have the options to set USE_DOCKER_WITH_CFTLIB which will determine whether docker in tandem.
- Also includes gradle upgrade because this is needed for the latest cftlib version. Need to merge #2119 first.

```
[ ] Yes
[ ] No
```


build.gradle
```diff
@@ -479,6 +479,20 @@ task buildAllXlsx(type: Exec) {
...
 
+task loadEnvSecrets() {
+  doLast {
+    if (!project.file('./.aat-env').exists()) {
+      new ByteArrayOutputStream().withStream { os ->
+        exec {
+          commandLine 'az', 'keyvault', 'secret', 'show', '--vault-name', 'probate-aat', '-o', 'tsv', '--query', 'value', '--name', 'probate-bo-dot-env'
+          standardOutput = os
+        }
+        project.file('./.aat-env').write(new String(os.toString().replace('\n', '').decodeBase64(), java.nio.charset.StandardCharsets.UTF_8))
```
DelythJustice: Not quite sure what this is for but it looks like if it gets updated in the vault then the old local copy will overide it. Which might not be what's needed?

danielbe11: My thinking was that we want to give the dev the chance to retain the old env file incase theres a problem with the new secret, but I think you're right, we should be fine to overwrite the env file each time the task is run - the vault keeps a history of old secrets in case something goes wrong.

sanjayparekh-hmcts (approved):

FeliTam (approved):

PR #2120, opened 2023-01-11
Dtspb 3116 new po box and sol scan

DelythJustice:
DTSPB-3116

### Change description ###

```
[ ] Yes
[x] No
```


DelythJustice:
This was investigation, we went with having exela set the po box instead of hardcoding for sols

PR #2146, opened 2023-01-20
DTSPB-3242

nilay913:
https://tools.hmcts.net/jira/browse/DTSPB-3242

### Change description ###
Added field documentUploadedAfterCaseStopped to CaseData, updated evidenceAdded in documentController for functionality. 
If case is ongoing, then uploading documents will always update last evidence added date.
If case is stopped, then:
If documentUploadedAfterCaseStopped = "No", documentUploadedAfterCaseStopped is set to "Yes", and last evidence added date is updated.
If documentUploadedAfterCaseStopped = "Yes", last evidence added date is not updated.
When case is reinstated, documentUploadedAfterCaseStopped is set to "No".

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/probate/controller/DocumentController.java
```diff
@@ -354,7 +354,20 @@ public ResponseEntity<CallbackResponse> reprint(@RequestBody CallbackRequest cal
 
     @PostMapping(path = "/evidenceAdded", consumes = MediaType.APPLICATION_JSON_VALUE)
     public ResponseEntity<CallbackResponse> evidenceAdded(@RequestBody CallbackRequest callbackRequest) {
-        evidenceUploadService.updateLastEvidenceAddedDate(callbackRequest.getCaseDetails());
+        CaseDetails caseDetails = callbackRequest.getCaseDetails();
+        CaseData caseData = caseDetails.getData();
```
DelythJustice, line 370: I think this won't satisfy scenario 3

src/main/java/uk/gov/hmcts/probate/controller/DocumentController.java
```diff
@@ -354,7 +354,21 @@ public ResponseEntity<CallbackResponse> reprint(@RequestBody CallbackRequest cal
...
+        if (caseDetails.getState().equalsIgnoreCase("BOCaseStopped")) {
+            log.info("Case is stopped: {} ", caseDetails.getId());
+            if (caseData.getDocumentUploadedAfterCaseStopped() != null
```
sanjayparekh-hmcts, line 374: Did we talk about the addition of this new field documentUploadedAfterCaseStopped, needing a similar update in probate-commons?

src/main/java/uk/gov/hmcts/probate/controller/DocumentController.java
```diff
@@ -354,7 +354,21 @@ public ResponseEntity<CallbackResponse> reprint(@RequestBody CallbackRequest cal
 
     @PostMapping(path = "/evidenceAdded", consumes = MediaType.APPLICATION_JSON_VALUE)
     public ResponseEntity<CallbackResponse> evidenceAdded(@RequestBody CallbackRequest callbackRequest) {
-        evidenceUploadService.updateLastEvidenceAddedDate(callbackRequest.getCaseDetails());
+        CaseDetails caseDetails = callbackRequest.getCaseDetails();
+        CaseData caseData = caseDetails.getData();
+        if (caseDetails.getState().equalsIgnoreCase("BOCaseStopped")) {
```
DelythJustice: I think we use STATE_BO_CASE_STOPPED from StateConstants

src/main/java/uk/gov/hmcts/probate/controller/DocumentController.java
nilay913: https://github.com/hmcts/probate-commons/pull/150/files
let me know if its lacking / if more is needed in it

src/main/java/uk/gov/hmcts/probate/controller/DocumentController.java
nilay913: the caseDetails.getState() returns the actual State ID saved in the data, is there another way I should be calling it?

DelythJustice: Not sure what you are asking, we have a value defined for it in the code so you could use that instead of hardcoded in quotes? Not that it's likely to change tbh

DelythJustice:
My thoughts on this are that we still might end up with the new flag set to 'yes' when the code is not expecting it due to our need to make change requests which only fix the state.
I think setting it to 'no' in the else loop (not stopped) would be one way to sort it.

DelythJustice (approved):

sanjayparekh-hmcts (approved):

src/test/resources/stoppedCaseAfterUploadingDocuments.json
```diff
@@ -0,0 +1,242 @@
+{
```
sanjayparekh-hmcts: maybe move these into own dir?

src/main/java/uk/gov/hmcts/probate/service/CaseStoppedService.java
```diff
@@ -18,6 +18,7 @@ public class CaseStoppedService {
 
     public void caseStopped(CaseDetails caseDetails) {
         caseDetails.getData().setGrantStoppedDate(LocalDate.now());
+        caseDetails.getData().setDocumentUploadedAfterCaseStopped("No");
```
sanjayparekh-hmcts: get **No** from statics?

src/main/java/uk/gov/hmcts/probate/model/ccd/raw/response/ResponseCaseDataParent.java
```diff
@@ -87,4 +87,5 @@ public class ResponseCaseDataParent {
     protected String solsNeedsPBAPayment;
     protected OrganisationPolicy applicantOrganisationPolicy;
     protected LocalDate lastEvidenceAddedDate;
+    protected String documentUploadedAfterCaseStopped;
```
sanjayparekh-hmcts, line 90: is this actually saved down to CCD?
If so needs to be in probate-commons, thus there would be a build.gradle change

sanjayparekh-hmcts (changes requested):

src/test/resources/stoppedCaseAfterUploadingDocuments.json
nilay913: these files aren't actually needed anymore so deleted them.

sanjayparekh-hmcts (approved):

PR #2182, opened 2023-02-08
Dtspb 3202 Ordering of new scanned document types

nilay913:
https://tools.hmcts.net/jira/browse/DTSPB-3202
https://tools.hmcts.net/jira/browse/DTSPB-3154
https://tools.hmcts.net/jira/browse/DTSPB-3265

### Change description ###
Updated fixedLists with new scanned docTypes so they show up in CCD, and created an ordering service that is ran every time scanned docs are attached to a case or when a case is created from bulkscan that will order the document types as per the ticket. 
Also found and updated any code instances of Type:Other Subtype:Will scanned documents to allow Type:Will too, and updated tests accordingly. 

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/probate/service/DocumentOrderingService.java
```diff
@@ -0,0 +1,34 @@
...
+@AllArgsConstructor
+@Service
+public class DocumentOrderingService {
+    public void orderScannedDocuments(CallbackRequest callbackRequest) {
+        CaseDetails caseDetails = callbackRequest.getCaseDetails();
+        CaseData caseData = caseDetails.getData();
+        caseData.getScannedDocuments()
+            .sort((o1,o2) -> {
+                String[] orderOfDocs = new String[]{"coversheet", "form", "will", "forensic_sheets",
+                    "supporting_documents", "iht", "pps_legal_statement", "cherished", "other"};
```
DelythJustice: think we could have this set somewhere so we don't recreate it every call?

DelythJustice: I know there won't be loads of entries but did you consider an enum for these and use order in enum for the comparator?

src/main/java/uk/gov/hmcts/probate/service/notification/GrantOfRepresentationPersonalisationService.java
```diff
@@ -127,8 +129,10 @@ private String parseDelimiters(String value) {
 
     private String getWillReferenceNumber(CaseData data) {
         for (CollectionMember<ScannedDocument> document : data.getScannedDocuments()) {
-            if (document.getValue().getSubtype() != null && document.getValue().getSubtype()
-                .equalsIgnoreCase(DOC_SUBTYPE_WILL)) {
+            if (document.getValue().getSubtype() != null
```
DelythJustice: do you still want to check the subtype is not null here even if looking only at type will in the or?

src/main/java/uk/gov/hmcts/probate/service/DocumentOrderingService.java
nilay913: yep added the set and the relevant doc types in the constants file, main reason I used the for loop was to get where it should belong, not sure if the enum method would be more efficient or not

src/main/java/uk/gov/hmcts/probate/service/notification/GrantOfRepresentationPersonalisationService.java
nilay913: good point, left it in as it was there before but thinking about it, in case 1 its checked subtype is will already and case 2 there shouldn't be any subtype nulls with type will, will see if can remove it

src/main/java/uk/gov/hmcts/probate/controller/NotificationController.java
```diff
@@ -213,6 +215,10 @@ public ResponseEntity<CallbackResponse> redeclarationSot(@RequestBody CallbackRe
     public ResponseEntity<CallbackResponse> sendGrantReceivedNotification(
         @RequestBody CallbackRequest callbackRequest) throws NotificationClientException {
         caseDataTransformer.transformCaseDataForEvidenceHandledForCreateBulkscan(callbackRequest);
+        if (callbackRequest.getCaseDetails().getData().getScannedDocuments() != null) {
```
sanjayparekh-hmcts: I thought you were gonna get rid of this null check here?

src/main/java/uk/gov/hmcts/probate/controller/NotificationController.java
```diff
@@ -230,6 +236,9 @@ public ResponseEntity<CallbackResponse> startDelayedNotificationPeriod(
         caseDataTransformer.transformCaseDataForAttachDocuments(callbackRequest);
         evidenceUploadService.updateLastEvidenceAddedDate(callbackRequest.getCaseDetails());
         CaseData caseData = callbackRequest.getCaseDetails().getData();
+        if (caseData.getScannedDocuments() != null) {
```
sanjayparekh-hmcts: and here

src/main/java/uk/gov/hmcts/probate/model/Constants.java
```diff
@@ -142,6 +153,17 @@ public static List<String> getSuccessorTitleClearingTypes() {
     public static final String EMAIL_REGEX = "[a-zA-Z0-9#$%'+=?^_`{|}~-]{1,30}(?:\\.[^.\\n]{1,30})"
         + "{0,30}@[a-zA-Z0-9](?:[a-zA-Z0-9-.]{0,30}[a-zA-Z0-9])?\\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,10}[a-zA-Z0-9])?";
 
+    public static final List<String> orderOfDocs = new ArrayList<>(asList(
```
sanjayparekh-hmcts: static name in CAPs please

src/main/java/uk/gov/hmcts/probate/service/ReprintService.java
```diff
@@ -27,7 +27,8 @@ public class ReprintService {
     public static final String LABEL_REISSUED_GRANT = "ReissuedGrant";
     public static final String LABEL_WILL = "Will";
     public static final String LABEL_SOT = "SOT";
-    public static final String WILL_DOC_TYPE = "Other";
+    public static final String WILL_DOC_TYPE = "will";
```
sanjayparekh-hmcts: are these actually the same values that you have set up as statics in Constants - reuse those?

src/main/java/uk/gov/hmcts/probate/service/ScannedDocumentOrderingService.java
```diff
@@ -0,0 +1,38 @@
...
+import uk.gov.hmcts.probate.model.Constants;
+
+import java.util.List;
+
+@Slf4j
+@AllArgsConstructor
+@Service
+public class ScannedDocumentOrderingService {
+    public void orderScannedDocuments(List<CollectionMember<ScannedDocument>> scannedDocs) {
+        scannedDocs
```
sanjayparekh-hmcts: put the null check here

src/test/resources/expectedOrderedScannedDocuments.json
```diff
@@ -0,0 +1,169 @@
+{
```
sanjayparekh-hmcts: maybe we should put these json's into their own directory?

src/test/resources/unorderedScannedDocuments.json
```diff
@@ -0,0 +1,169 @@
+{
```
sanjayparekh-hmcts: these too

sanjayparekh-hmcts (changes requested):

sanjayparekh-hmcts (approved):

PR #2277, opened 2023-04-11
DTSPB-3449 Post Grant issue changes

IswaryaPepakayala:

https://tools.hmcts.net/jira/browse/DTSPB-3449

### Change description ###

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/probate/service/CaseQueryService.java
```diff
@@ -89,8 +89,14 @@ private static <T> T nonNull(@Nullable T result) {
 
     public List<ReturnedCaseDetails> findGrantIssuedCasesWithGrantIssuedDate(String invokedFrom, String queryDate) {
         BoolQueryBuilder query = boolQuery();
+        BoolQueryBuilder stateChecks = boolQuery();
 
-        query.must(matchQuery(STATE, STATE_MATCH));
+        for (String stateToMatch : Arrays.asList(STATE_MATCH)) {
```
DelythJustice, line 94: I think there's term query would be faster that match (cos it will be looking at variants not just exact) but maybe we aren't currently using term query?
eg QueryBuilders.termQuery as opposed to QueryBuilders.matchQuery (which does the same as your new MatchQueryBuilder)

DelythJustice: Probably something for the back burner - I only see these match queries currently, they might well be mapped as text which needs match :)

DelythJustice (approved):

src/main/java/uk/gov/hmcts/probate/service/CaseQueryService.java
IswaryaPepakayala: Got this from google
 termQuery searches for an exact match of the specified term in the specified field, without performing any analysis on the search term. 
if you are searching for text data such as titles or descriptions, and want to allow for variations in the search term such as typos or synonyms, then matchQuery is likely the better option. This is because matchQuery analyzes the search term and returns documents that contain any of the terms produced by the analysis, making it more flexible and capable of handling variations in the search term.

PR #2362, opened 2023-08-10
Try change collation

timja:
https://github.com/hashicorp/terraform-provider-azurerm/pull/22689

> Error: collation contains unknown collation en-GB
with module.db-v11.azurerm_postgresql_database.postgres-db,
on .terraform/modules/db-v11/main.tf line 58, in resource "azurerm_postgresql_database" "postgres-db":


timja:
can't do this as it forces replacement

PR #2364, opened 2023-08-14
DTSPB-3737 fix missing deceased alias

FeliTam:
https://tools.hmcts.net/jira/browse/DTSPB-3737

### Change description ###
Fix missing deceased alias

```
[ ] Yes
[ x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/probate/transformer/CallbackResponseTransformer.java
+++ b/src/main/java/uk/gov/hmcts/probate/transformer/CallbackResponseTransformer.java
@@ -1459,7 +1459,7 @@ public class CallbackResponseTransformer {
                     .map(CollectionMember::getValue)
                     .map(this::buildDeceasedAliasNameExecutor)
                     .map(alias -> new CollectionMember<>(null, alias))
-                    .collect(Collectors.toList());
+                    .toList();
         }
         if (deceasedAliasNames.isEmpty()) {
             builder
--- a/src/main/java/uk/gov/hmcts/probate/transformer/CallbackResponseTransformer.java
+++ b/src/main/java/uk/gov/hmcts/probate/transformer/CallbackResponseTransformer.java
@@ -1450,8 +1450,25 @@ public class CallbackResponseTransformer {
         builder
                 .ihtReferenceNumber(caseData.getIhtReferenceNumber())
                 .primaryApplicantAlias(caseData.getPrimaryApplicantAlias())
-                .solsExecutorAliasNames(caseData.getSolsExecutorAliasNames())
-                .solsDeceasedAliasNamesList(caseData.getSolsDeceasedAliasNamesList());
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/probate/transformer/CallbackResponseTransformer.java
```diff
@@ -1450,8 +1450,25 @@ private void updateCaseBuilderForTransformCase(CaseData caseData, ResponseCaseDa
         builder
                 .ihtReferenceNumber(caseData.getIhtReferenceNumber())
                 .primaryApplicantAlias(caseData.getPrimaryApplicantAlias())
-                .solsExecutorAliasNames(caseData.getSolsExecutorAliasNames())
-                .solsDeceasedAliasNamesList(caseData.getSolsDeceasedAliasNamesList());
+                .solsExecutorAliasNames(caseData.getSolsExecutorAliasNames());
+
+        List<CollectionMember<AliasName>> deceasedAliasNames = EMPTY_LIST;
```
DelythJustice, line 1455: think you can do:

List<CollectionMember<AliasName>> deceasedAliasNames = Optional.ofNullable(caseData.getDeceasedAliasNameList())
        .stream() etc then not need the null check

src/main/java/uk/gov/hmcts/probate/transformer/CallbackResponseTransformer.java
```diff
@@ -1450,8 +1450,25 @@ private void updateCaseBuilderForTransformCase(CaseData caseData, ResponseCaseDa
...
+                    .toList();
+        }
+        if (deceasedAliasNames.isEmpty()) {
+            builder
+                    .solsDeceasedAliasNamesList(caseData.getSolsDeceasedAliasNamesList());
+        } else {
+            builder
+                    .solsDeceasedAliasNamesList(deceasedAliasNames)
+                    .deceasedAliasNamesList(null);
+        }
```
DelythJustice, line 1471: Is this what is needed? As in if there's values in the deceasedAliasNameList then solsDeceasedAliasNamesList is filled with deceasedAliasNameList. Wondering if there could be values in solsDeseasedAliasNamesList already?

FeliTam: This block is from https://github.com/hmcts/probate-back-office/blob/master/src/main/java/uk/gov/hmcts/probate/transformer/CallbackResponseTransformer.java#L1425-L1441
Think the logic is to copy deceasedAliasNamesList(only accessible from PA journey) to solsDeceasedAliasNamesList, and then deceasedAliasNamesList will be set to empty.
In case matching, only solsDeceasedAliasNamesList is included as search criteria.

DelythJustice (approved):

PR #2500, opened 2024-02-22
Dtspb 3936 channel choice notifications

GEvans-HMCTS:
Resolves: https://tools.hmcts.net/jira/browse/DTSPB-3936


FeliTam (commented):
ChannelChoice is created to replace paperForm field,
can you also remove paperForm from both NotificationService and TemplateService?

src/main/java/uk/gov/hmcts/probate/service/NotificationService.java
```diff
@@ -112,8 +112,8 @@ public Document sendEmail(State state, CaseDetails caseDetails, Optional<CaseOri
             caseData.getPaperForm(), caseDetails.getId(),
```
FeliTam: can you replace the paperForm with channelChoice in this log.info block as well

FeliTam (approved):

PR #2502, opened 2024-03-04
DTSPB-3959 Caveat payment Journey PP changes

IswaryaPepakayala:

https://tools.hmcts.net/jira/browse/DTSPB-3959

### Change description ###

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/probate/validator/CaveatAcknowledgementValidationRule.java
```diff
@@ -0,0 +1,39 @@
...
+    public static final String PAYMENT_CONFIRMATION = "paymentAcknowledgement";
+
+    private final BusinessValidationMessageRetriever businessValidationMessageRetriever;
+
+    public void validate(CaveatDetails caseDetails) {
+        CaveatData caveatData = caseDetails.getData();
+        List<String> paymentConfirmation = caveatData.getPaymentConfirmCheckbox();
+        String confirmation = (null != paymentConfirmation && !paymentConfirmation.isEmpty())
+                ? caveatData.getPaymentConfirmCheckbox().get(0) : "";
+        log.info("Payment confirmation: {}", confirmation);
```
FeliTam: think we can remove this log, as the value is either paymentAcknowledgement or empty string

src/main/java/uk/gov/hmcts/probate/validator/CaveatAcknowledgementValidationRule.java
```diff
@@ -0,0 +1,39 @@
...
+        if (!PAYMENT_CONFIRMATION.equals(confirmation)) {
+            String userMessage = businessValidationMessageRetriever
+                    .getMessage(PAYMENT_ACKNOWLEDGEMENT, null, Locale.UK);
+            throw new BusinessValidationException(userMessage,
+                    "You must confirm that you understand that you must complete payment after you've submitted your "
+                            + "application");
```
FeliTam: this message will be in the log only, don't need to be the same as the one in UI
maybe log the case id here

src/main/java/uk/gov/hmcts/probate/validator/CaveatAcknowledgementValidationRule.java
IswaryaPepakayala: Sorry kept this for testing and forgot to remove it.Removed now

src/main/java/uk/gov/hmcts/probate/validator/CaveatAcknowledgementValidationRule.java
IswaryaPepakayala: Changed it

FeliTam (dismissed):

MarcusHMCTS (approved):

PR #2867, opened 2024-11-11
DTSPB-4242 search for caseworker author name

FeliTam:
DTSPB-4242

### Change description ###
Enable search for caseworker author name

```
[ ] Yes
[x] No
```


tom-saunders-cts (commented):
would we benefit from clearer interfaces by making `authToken` a typed entity and encapsulating the needed interface to idam etc within? means we could mock out the behaviours slightly cleaner and maybe makes the method signatures clearer at a glance: `doWhatever(String, String)` vs `doWhatever(Auth, State)`

src/main/java/uk/gov/hmcts/probate/controller/BusinessValidationController.java
```diff
@@ -375,7 +378,7 @@ public ResponseEntity<CallbackResponse> validateCheckListDetails(
             eventValidationService.validateRequest(callbackRequest, checkListAmendCaseValidationRules);
 
         if (response.getErrors().isEmpty()) {
-            response = callbackResponseTransformer.selectForQA(callbackRequest);
+            response = callbackResponseTransformer.selectForQA(callbackRequest, authToken);
```
IswaryaPepakayala: This is an obsolete event.so do we need to change this?

FeliTam: removed related code

MarcusHMCTS (dismissed):

MarcusHMCTS (approved):

PR #2870, opened 2024-11-18
Dtspb 4329 reasons for registrar escalation

MarcusHMCTS:
[DTSPB-4329](https://tools.hmcts.net/jira/browse/DTSPB-4329)

### Change description ###

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/probate/model/ccd/raw/response/ResponseCaseDataParent.java
```diff
@@ -75,6 +75,7 @@ public class ResponseCaseDataParent {
     protected final String solsReviewSOTConfirmCheckbox2Names;
     protected final String taskList;
     protected final String escalatedDate;
+    protected final String registrarEscalateReason;
```
tom-saunders-cts: Could/should this be an enum type rather than a String if we have a defined number of values it is expected to be? (e.g. similar to how `applicationType` is handled at [line 346](https://github.com/hmcts/probate-back-office/blob/6f8881e13757efd7b3c4dfc6aa88a874a13943b8/src/main/java/uk/gov/hmcts/probate/model/ccd/raw/request/CaseData.java#L346))

tom-saunders-cts: This was intended to be put on the CaseData class, but I think the same is true here

MarcusHMCTS: registrarEscalateReason come from the [FixedList.json](https://github.com/hmcts/probate-back-office/pull/2870/files#diff-e5c6156ba9a8706ce144242f12c52925cbfc6639fddece01a4c79431b2e5f948R424-R425)

MarcusHMCTS: Enum added and moved to CaseData, thanks

tom-saunders-cts (dismissed):

IswaryaPepakayala (approved):

PR #2882, opened 2024-11-20
DTSPB-4293 Added Ad Colligenda Bona case type changes

IswaryaPepakayala:

https://tools.hmcts.net/jira/browse/DTSPB-4293

### Change description ###

```
[ ] Yes
[X] No
```


MarcusHMCTS (dismissed):

src/functionalTest/java/uk/gov/hmcts/probate/functional/documents/GrantGenerationTests.java
```diff
@@ -438,6 +454,40 @@ void verifyGenerateAllEnglishIntestacySolicitorGrantTypesWhenDeceasedDomiciledNo
...
+        String response = generateGrantDocument(adColligendaPayload + AD_COLLIGENDA_JSON,
+                GENERATE_GRANT_DRAFT);
+        assertTrue(response.contains(AD_COLLIGENDA_GRANT_TEXT));
+
+        response = generateGrantDocument(adColligendaPayload + AD_COLLIGENDA_JSON, GENERATE_GRANT);
+        assertTrue(response.contains(AD_COLLIGENDA_GRANT_TEXT));
+
+        response = generateGrantDocument(adColligendaPayload + AD_COLLIGENDA_REISSUE_JSON,
+                GENERATE_GRANT_REISSUE);
+        assertTrue(response.contains(AD_COLLIGENDA_GRANT_TEXT));
```
tom-saunders-cts: i'd argue this is either three distinct test which we want to be able to pass/fail distinctly or we should be using `assertAll` to gather the failures so they are all reported as needed. 

i'd also suggest that we should have messaging for the `assertTrue` calls here as otherwise the test will simply report a message along the lines of `expected <true> but was: <false> at ${failing_line}` which is not helpful to someone looking through the logs to determine _why_ the test failed

src/integrationTest/java/uk/gov/hmcts/probate/service/docmosis/DocumentTemplateServiceIT.java
```diff
@@ -174,4 +174,47 @@ void shouldGetAdmonWillTemplate() {
         assertEquals(DocumentType.WELSH_ADMON_WILL_GRANT_REISSUE_DRAFT, responseReissueWelshDraft);
 
     }
+
+    @Test
+    void shouldGetAdColligendaBonaTemplate() {
```
tom-saunders-cts, line 179: again this feels like either these tests are independent or should use `assertAll`. The exception message in this case are a little better (expected X, got Y) but if we do have multiple failures it might help to explain what each is.

src/main/java/uk/gov/hmcts/probate/controller/BusinessValidationController.java
```diff
@@ -578,6 +580,7 @@ public ResponseEntity<CallbackResponse> paperFormCaseDetails(
         caseDataTransformer.transformCaseDataForPaperForm(callbackRequest);
         handOffLegacyTransformer.setHandOffToLegacySiteYes(callbackRequest);
         validateForPayloadErrors(callbackRequest, bindingResult);
+        caseTypeValidationRule.validate(callbackRequest.getCaseDetails());
```
tom-saunders-cts: should this validation be performed anywhere else? we're only preventing manual paper applications?

src/main/java/uk/gov/hmcts/probate/validator/CaseTypeValidationRule.java
```diff
@@ -0,0 +1,29 @@
...
+import uk.gov.hmcts.probate.model.ccd.raw.request.CaseData;
+import uk.gov.hmcts.probate.model.ccd.raw.request.CaseDetails;
+import uk.gov.hmcts.probate.service.BusinessValidationMessageRetriever;
+
+import java.util.Locale;
+
+
+@Component
+@RequiredArgsConstructor
+public class CaseTypeValidationRule {
```
tom-saunders-cts: knowing what this does (fail if the case type is Ad Collegenda Bona and nothing else) is this named appropriately? (Obviously naming things is difficult, but this sort of suggests that it's a generic validation of case types)

src/main/java/uk/gov/hmcts/probate/service/BulkPrintService.java
```diff
@@ -207,15 +211,20 @@ private List<uk.gov.hmcts.reform.sendletter.api.model.v3.Document> arrangePdfDoc
         if (documentTransformer.hasDocumentWithType(Arrays.asList(grantDocument), DIGITAL_GRANT)
             || documentTransformer.hasDocumentWithType(Arrays.asList(grantDocument), ADMON_WILL_GRANT)
             || documentTransformer.hasDocumentWithType(Arrays.asList(grantDocument), INTESTACY_GRANT)
+            || documentTransformer.hasDocumentWithType(Arrays.asList(grantDocument), AD_COLLIGENDA_BONA_GRANT)
```
tom-saunders-cts: This feels like we're repeating ourselves a lot

src/main/java/uk/gov/hmcts/probate/service/template/pdf/PDFDecoratorService.java
```diff
@@ -97,6 +99,8 @@ private String addExtraCaseData(String dataJson, Object data, DocumentType docum
                     || documentType.equals(INTESTACY_GRANT_DRAFT)
                     || documentType.equals(ADMON_WILL_GRANT)
                     || documentType.equals(ADMON_WILL_GRANT_DRAFT)
+                    || documentType.equals(AD_COLLIGENDA_BONA_GRANT)
+                    || documentType.equals(AD_COLLIGENDA_BONA_GRANT_DRAFT)
```
tom-saunders-cts, line 103: Feels like we're doing this adding grant types to logic in a lot of places - can/should we have common definitions of 
``` java 
Collection<DocumentType> GRANT_DOC_TYPES = Set.of(
        DIGITAL_GRANT,
        DIGITAL_GRANT_DRAFT,
        // etc.
```
and then do something like:
``` java
} else if (GRANT_DOC_TYPES.contains(documentType)) {
```

src/main/java/uk/gov/hmcts/probate/service/template/pdf/PDFManagementService.java
```diff
@@ -75,6 +75,11 @@ public Document generateAndUpload(CallbackRequest callbackRequest, DocumentType
                     .setGrantSignatureBase64(decryptedFileAsBase64String(pdfServiceConfiguration
                         .getGrantSignatureEncryptedFile()));
                 break;
+            case AD_COLLIGENDA_BONA_GRANT:
+                callbackRequest.getCaseDetails()
+                        .setGrantSignatureBase64(decryptedFileAsBase64String(pdfServiceConfiguration
+                                .getGrantSignatureEncryptedFile()));
+                break;
```
tom-saunders-cts, line 82: This branch seems identical to the previous three? would doing similar to in the PDFDecoratorService be a better approach if the behaviour is always X or Y:
``` java
// obviously FINAL_GRANT_TYPES doesn't seem to exist currently
if (FINAL_GRANT_TYPES.contains(documentType)) {
    // signature handling
}
```

The below is correct, but this field is only actually used for the document generation (is not persisted?) and holds a base64 encoded image of a signature, rather than being a digital signature of the document itself
~Further, is this handling correct? It seems like we are:~
1. ~getting an encrypted file~
2. ~decrypting it~
3. ~converting the binary file to base64~
4. ~storing that base64 as the signature~

~This is not what would commonly be called a signature - it's the entire file and we've removed any protection having it stored encrypted provided us?~

src/main/java/uk/gov/hmcts/probate/transformer/DocumentTransformer.java
```diff
@@ -25,49 +25,27 @@ public boolean hasDocumentWithType(List<Document> documents, DocumentType docume
...
+                    INTESTACY_GRANT, INTESTACY_GRANT_REISSUE_DRAFT, INTESTACY_GRANT_REISSUE,
+                    WELSH_INTESTACY_GRANT_REISSUE_DRAFT, WELSH_INTESTACY_GRANT_REISSUE, ADMON_WILL_GRANT_DRAFT,
+                    ADMON_WILL_GRANT, ADMON_WILL_GRANT_REISSUE_DRAFT, ADMON_WILL_GRANT_REISSUE,
+                    WELSH_ADMON_WILL_GRANT_REISSUE_DRAFT, WELSH_ADMON_WILL_GRANT_REISSUE,
+                    AD_COLLIGENDA_BONA_GRANT_DRAFT, AD_COLLIGENDA_BONA_GRANT, AD_COLLIGENDA_BONA_GRANT_REISSUE_DRAFT,
+                    AD_COLLIGENDA_BONA_GRANT_REISSUE, WELSH_AD_COLLIGENDA_BONA_GRANT_REISSUE_DRAFT,
+                    WELSH_AD_COLLIGENDA_BONA_GRANT_REISSUE, SOT_INFORMATION_REQUEST, GRANT_COVER, ASSEMBLED_LETTER,
+                    BLANK_LETTER, WELSH_DIGITAL_GRANT_DRAFT, WELSH_DIGITAL_GRANT, WELSH_ADMON_WILL_GRANT_DRAFT,
+                    WELSH_ADMON_WILL_GRANT, WELSH_INTESTACY_GRANT_DRAFT, WELSH_INTESTACY_GRANT,
+                    WELSH_AD_COLLIGENDA_BONA_GRANT_DRAFT, WELSH_AD_COLLIGENDA_BONA_GRANT:
```
tom-saunders-cts: the before here seems much clearer to read? (though again, would having a common collection which we can do `if (COMMON.contains(documentType)) {` be further cleaner?

src/main/java/uk/gov/hmcts/probate/transformer/ReprintTransformer.java
```diff
@@ -85,24 +85,16 @@ private boolean isFromScannedDOcuments(ScannedDocument document) {
...
+        return switch (document.getDocumentType()) {
+            case DIGITAL_GRANT, INTESTACY_GRANT, ADMON_WILL_GRANT, AD_COLLIGENDA_BONA_GRANT, WELSH_DIGITAL_GRANT,
+                    WELSH_INTESTACY_GRANT, WELSH_ADMON_WILL_GRANT, WELSH_AD_COLLIGENDA_BONA_GRANT ->
+                    true;
+            case DIGITAL_GRANT_REISSUE, INTESTACY_GRANT_REISSUE, ADMON_WILL_GRANT_REISSUE,
+                    AD_COLLIGENDA_BONA_GRANT_REISSUE ->
+                    true;
+            case STATEMENT_OF_TRUTH, WELSH_STATEMENT_OF_TRUTH -> true;
+            default -> false;
+        };
```
tom-saunders-cts: again we have multiple identical branches which boil down to `return SOME_COLLECTION.contains(documentType)`

src/main/resources/templates/pdf/adColligendaBonaGrant.html
tom-saunders-cts, line 1: We have this template for the new grant, and also have updated the docmosis templates for all grant types.

Which one is actually used?

src/test/java/uk/gov/hmcts/probate/validator/CaseTypeValidationRuleTest.java
```diff
@@ -0,0 +1,56 @@
...
+    @InjectMocks
+    private CaseTypeValidationRule caseTypeValidationRule;
+
+    @Mock
+    private BusinessValidationMessageRetriever businessValidationMessageRetriever;
+    private static final String[] LAST_MODIFIED = {"2018", "1", "1", "0", "0", "0", "0"};
+    private static final Long CASE_ID = 12345678987654321L;
+    private CaseData dataMock;
+    @Mock
+    private CaseDetails detailsMock;
```
tom-saunders-cts: Should this be a class level member - it's also never used as a mock? (ditto the `CaseData` above)

src/test/java/uk/gov/hmcts/probate/validator/CaseTypeValidationRuleTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @BeforeEach
+    public void setUp() {
+        MockitoAnnotations.openMocks(this);
+        dataMock = CaseData.builder().caseType("gop").build();
+        detailsMock = new CaseDetails(dataMock, LAST_MODIFIED, CASE_ID);
+    }
```
tom-saunders-cts: It's not always needed, but not closing the mocks we open can cause problems in some situations:
```suggestion
    private AutoCloseable closeableMocks;
    @BeforeEach
    public void setUp() {
        closeableMocks = MockitoAnnotations.openMocks(this);
        dataMock = CaseData.builder().caseType("gop").build();
        detailsMock = new CaseDetails(dataMock, LAST_MODIFIED, CASE_ID);
    }
    
    @AfterEach
    void cleanUp() throws Exception {
        closeableMocks.close();
    }
```

src/main/resources/templates/pdf/adColligendaBonaGrant.html
IswaryaPepakayala: This is for Grant document and Docmosis templates are for Reissue and Welsh Documents

src/main/java/uk/gov/hmcts/probate/transformer/ReprintTransformer.java
IswaryaPepakayala: Modified

src/main/java/uk/gov/hmcts/probate/transformer/DocumentTransformer.java
IswaryaPepakayala: Modified

src/test/java/uk/gov/hmcts/probate/validator/CaseTypeValidationRuleTest.java
IswaryaPepakayala: Updated

src/main/java/uk/gov/hmcts/probate/service/BulkPrintService.java
IswaryaPepakayala: Simplified

src/main/java/uk/gov/hmcts/probate/validator/CaseTypeValidationRule.java
IswaryaPepakayala: Changed name of the class

src/main/java/uk/gov/hmcts/probate/controller/BusinessValidationController.java
IswaryaPepakayala: Yes that is enough

src/integrationTest/java/uk/gov/hmcts/probate/service/docmosis/DocumentTemplateServiceIT.java
IswaryaPepakayala: Modified

src/test/java/uk/gov/hmcts/probate/validator/CaseTypeValidationRuleTest.java
IswaryaPepakayala: Removed mock

src/functionalTest/java/uk/gov/hmcts/probate/functional/documents/GrantGenerationTests.java
IswaryaPepakayala: Added as separate ones

MarcusHMCTS (approved):

PR #2944, opened 2025-01-03
DTSPB-4395 added new template id to caveat 

abbasiTahir:

https://tools.hmcts.net/jira/browse/DTSPB-4395

### Change description ###

```
[ ] Yes
[x ] No
```


src/main/resources/application.properties
tom-saunders-cts, line 1: should these changes have been included?

src/main/resources/application.yml
```diff
@@ -246,7 +246,7 @@ notifications:
           applicationReceivedPaperFormCaseworker: "01bfc52c-0850-4155-ae03-e6d638286288"
           documentReceived: "24161fba-65f5-4384-9c4f-ef75f4a5c666"
           caseStopped: "60b95b7c-0a76-4608-b549-b3a579cfa3d0"
-          caseStoppedCaveat: "937291a0-c0ae-40a4-a8cf-46a3452faff0"
+          caseStoppedCaveat: "d5084257-f475-415a-b7d0-2427b5c580b1"
```
tom-saunders-cts, line 261: - old https://www.notifications.service.gov.uk/services/4809d6a8-a10b-405d-88eb-1fee9b3042e3/templates/937291a0-c0ae-40a4-a8cf-46a3452faff0
- new https://www.notifications.service.gov.uk/services/4809d6a8-a10b-405d-88eb-1fee9b3042e3/templates/d5084257-f475-415a-b7d0-2427b5c580b1

new appears to have "We won't be able to give you legal advice." twice

src/main/resources/application.yml
```diff
@@ -297,7 +297,7 @@ notifications:
           applicationReceivedPaperFormCaseworker: "22c81233-a213-4520-b6bc-896fe8f6988a"
           documentReceived: "39b7cef5-8202-479e-9f1e-86a76f964221"
           caseStopped: "80b576b4-1a33-4920-837a-f2904784c9e2"
-          caseStoppedCaveat: "1fde1860-b827-45e7-b430-1406f74beb93"
+          caseStoppedCaveat: "4aec6fa3-7e76-406a-9953-2f12d5d29f81"
```
tom-saunders-cts, line 312: - old https://www.notifications.service.gov.uk/services/4809d6a8-a10b-405d-88eb-1fee9b3042e3/templates/1fde1860-b827-45e7-b430-1406f74beb93
- new https://www.notifications.service.gov.uk/services/4809d6a8-a10b-405d-88eb-1fee9b3042e3/templates/4aec6fa3-7e76-406a-9953-2f12d5d29f81

new appears to only have changed the text in english, and the modified text doesn't match what is in the ticket

src/main/resources/application.yml
abbasiTahir: resolved

src/main/resources/application.yml
abbasiTahir: resolved

tom-saunders-cts: there are some other weird handling parts of these messages, but i think we can handle those outside the scope of the PR(?) - i have added a comment in the ticket to flag them and we can follow up through that since we can update the template without a PR

tom-saunders-cts (dismissed):
the alternative would be to directly update the existing template and make no change, but i think this is probably the better solution as we can track and control when that change goes live

src/main/resources/application.yml
abbasiTahir: Thanks Tom good catch .

src/contractTest/resources/application.yml
```diff
@@ -286,7 +286,7 @@ notifications:
           applicationReceivedPaperFormCaseworker: "22c81233-a213-4520-b6bc-896fe8f6988a"
           documentReceived: "83f93f41-58a1-44f2-a9c4-d20d545b7528"
           caseStopped: "a767a0bd-4acc-413e-b510-5a71d24daf85"
-          caseStoppedCaveat: "937291a0-c0ae-40a4-a8cf-46a3452faff0"
+          caseStoppedCaveat: "d5084257-f475-415a-b7d0-2427b5c580b1"
```
FeliTam: this should use welsh version template

abbasiTahir: resolved

src/functionalTest/java/uk/gov/hmcts/probate/functional/notifications/SolBaCcdServiceNotificationTests.java
```diff
@@ -35,7 +35,7 @@ public class SolBaCcdServiceNotificationTests extends IntegrationTestBase {
     private static final String PAPER_FORM = "/case/paperForm";
 
     private static final String BIRMINGHAM_NO = "0300 303 0648";
-    private static final String REGISTRY_NO = "0113 389 6133";
+    private static final String REGISTRY_NO = " 0300 303 0648";
```
tom-saunders-cts, line 38: can we update the test(s) where this was failing to improve their failure messaging in case we have the same issue in the future? (given the error message was `org.opentest4j.AssertionFailedError: expected: <true> but was: <false>` this doesn't actually provide any useful information on why it is failing.)

FeliTam (approved):

PR #2982, opened 2025-01-27
DTSPB-4471 update statement of truth in assemble letters

abbasiTahir:

-

https://tools.hmcts.net/jira/browse/DTSPB-4471

### Change description ###

```
[ ] Yes
[ x] No
```


src/main/resources/application.yml
```diff
@@ -113,9 +113,9 @@ docmosis:
     welshStatementOfTruth:
       templateName: FL-PRB-SOT-WEL-00236.doc
     assembledLetter:
-      templateName: FL-PRB-GNO-ENG-00133-DTSPB-2638.docx
+      templateName: FL-PRB-GNO-ENG-00237.docx
     blankLetter:
-      templateName: FL-PRB-GNO-ENG-01040.docx
+      templateName: FL-PRB-GNO-ENG-00237.docx.docx
```
tom-saunders-cts: i assume there is a reason these were separate documents 

we now have `.docx` duplicated

tom-saunders-cts (changes requested):

src/main/resources/application.yml
```diff
@@ -113,9 +113,9 @@ docmosis:
     welshStatementOfTruth:
       templateName: FL-PRB-SOT-WEL-00236.doc
     assembledLetter:
-      templateName: FL-PRB-GNO-ENG-00133-DTSPB-2638.docx
+      templateName: FL-PRB-GNO-ENG-00237.docx
     blankLetter:
-      templateName: FL-PRB-GNO-ENG-01040.docx
+      templateName: FL-PRB-GNO-ENG-00237.docx
```
tom-saunders-cts: is there not a reason these were different documents to start with? (i.e. `assembledLetter.templateName` and `blankLetter.templateName` were different before)

tom-saunders-cts (changes requested):

src/main/resources/application.yml
abbasiTahir: resolved

abbasiTahir: resolved .

tom-saunders-cts (dismissed):
I believe all the referenced templates are correct to what is requested in the ticket.

I suspect we may need to do some template tidying when this merges (so we aren't using `Template/Draft/... (copy)` in production)

jamiealbertelli:
> I believe all the referenced templates are correct to what is requested in the ticket.
> 
> I suspect we may need to do some template tidying when this merges (so we aren't using `Template/Draft/... (copy)` in production)

100% needed, Notify is a bit of a tip at the moment. Although the input needed from the service team may be large. Suppose we can compare existing template IDs with those in BO and go to them to confirm?

tom-saunders-cts (dismissed):

tom-saunders-cts (dismissed):

tom-saunders-cts (approved):

PR #3034, opened 2025-03-10
Dtspb 4557 remove pence value from smee and ford data

MarcusHMCTS:

[DTSPB-4557](https://tools.hmcts.net/jira/browse/DTSPB-4557)

### Change description ###

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/probate/service/notification/SmeeAndFordPersonalisationService.java
```diff
@@ -371,4 +371,11 @@ private String removeLastNewLine(String data) {
         return data.substring(0, data.lastIndexOf(NEW_LINE));
     }
 
+    private String getPoundValue(BigDecimal value) {
+        if (value == null || value.toString().length() < 2) {
+            return "0";
+        }
+        return value.toString().substring(0, value.toString().length() - 2);
+    }
+
```
tom-saunders-cts, line 380:
```suggestion
    private String getPoundValue(BigDecimal penceValue) {
        if (penceValue == null) {
            return "0";
        }
        final BigDecimal poundValue = penceValue.divideToIntegralValue(BigDecimal.valueOf(100L));
        return poundValue.toString();
    }
```

handles the (unexpected) case where we have fractional parts of a penny) and i think is slightly simpler to understand than taking a substring?

PR #3118, opened 2025-05-06
Dtspb 4580 stop reminder

MarcusHMCTS:

### Change description ###

```
[ ] Yes
[ ] No
```


src/main/java/uk/gov/hmcts/probate/schedule/SendNotificationsTask.java
```diff
@@ -0,0 +1,81 @@
...
+            } else {
+                log.info("Calling Send First Stop Reminder for date {}", firstStopReminderDate);
+                dataExtractDateValidator.dateValidator(firstStopReminderDate);
+                log.info("Perform Send First Stop Reminder started");
+                automatedNotificationService.sendNotification(firstStopReminderDate, FIRST_STOP_REMINDER);
+                log.info("Perform Send First Stop Reminder finished");
+
+            }
+        } catch (ApiClientException e) {
+            log.error(e.getMessage());
```
tom-saunders-cts: do we really just want to log the message (and not say the stack trace so we can work out what's happened)? similar question for the other catch & log instances - some of them may be reasonable, but if we're totally discarding the exception maybe not ideal.

this is especially important for the instance immediately below because that could be any Exception subclass, which might be signalling a failure we did not anticipate at all

src/main/java/uk/gov/hmcts/probate/service/notification/AutomatedNotificationService.java
```diff
@@ -0,0 +1,101 @@
...
+
+        List<Long> failedCases = new ArrayList<>();
+        securityUtils.setSecurityContextUserAsScheduler();
+        SecurityDTO securityDTO = securityUtils.getUserBySchedulerTokenAndServiceSecurityDTO();
+
+        try {
+            log.info("Fetch and process automated notification for tyoe: {}", type);
+            fetchAndProcessPages(strategy, date, securityDTO, failedCases);
+        } catch (Exception e) {
+            log.error("Error sending notifications for type {}: {}", type, e.getMessage(), e);
```
tom-saunders-cts: we're passing three parameters to a format string which consumes two - is the intention here to log the formatted message and the stack trace? if so we'll need to format the message separately and use the `error(String, Throwable)` method

src/main/java/uk/gov/hmcts/probate/model/ccd/raw/request/CaseData.java
```diff
@@ -631,6 +631,7 @@ public class CaseData extends CaseDataParent {
     private final List<CollectionMember<UploadDocument>> citizenDocumentsUploaded;
     private List<CollectionMember<CitizenResponse>> citizenResponses;
     private final String executorsNamed;
+    private final LocalDate firstStopReminderSentDate;
```
tom-saunders-cts: is this value persisted when other events occur on the case? should it be? specifically what happens if the case moves out of the stop state (for whatever reason) and then back in?

src/test/java/uk/gov/hmcts/probate/service/notification/SecondStopReminderNotificationTest.java
```diff
@@ -0,0 +1,46 @@
...
+import uk.gov.hmcts.probate.service.NotificationService;
+import uk.gov.hmcts.reform.ccd.client.model.CaseDetails;
+import uk.gov.service.notify.NotificationClientException;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(SpringExtension.class)
```
tom-saunders-cts: almost certainly do not need this

src/test/java/uk/gov/hmcts/probate/service/notification/FirstStopReminderNotificationTest.java
```diff
@@ -0,0 +1,46 @@
...
+import uk.gov.hmcts.probate.service.NotificationService;
+import uk.gov.hmcts.reform.ccd.client.model.CaseDetails;
+import uk.gov.service.notify.NotificationClientException;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(SpringExtension.class)
```
tom-saunders-cts: likely unneeded

src/main/java/uk/gov/hmcts/probate/schedule/SendNotificationsTask.java
```diff
@@ -0,0 +1,81 @@
...
+    @Value("${automated_notification.stop_reminder.second_notification_days}")
+    private int secondNotificationDays;
+
+    @Value("${adhocSchedulerJobDate}")
+    public String adHocJobDate;
+
+    @Override
+    public void run() {
+        log.info("Scheduled task SendNotificationsTask started");
+        String firstStopReminderDate = DATE_FORMAT.format(LocalDate.now().minusDays(firstNotificationDays));
```
tom-saunders-cts: should probably have a `Clock` and use `LocalDate.now(clock)` [docs](https://docs.oracle.com/javase/8/docs/api/java/time/LocalDate.html#now-java.time.Clock-) - means we can control time during testing

src/main/resources/application.yml
```diff
@@ -330,6 +336,10 @@ notifications:
           grantAwaitingDocumentation: "cc4aa69b-10fb-4a9c-9e0e-1851dfaa69e9"
           disposalReminder: "34fe5d9d-c57b-4bc2-9ceb-008e82cd5594"
           draftCasePaymentSuccess: "b881e8ea-76a5-412c-83d5-8e0f08791b16"
+          firstStopReminder: "3c4d8cae-2992-4950-95ab-2d1fd7e8ad7d"
+          firstStopReminderForHub: "a2856f90-39a7-4ff4-9107-13f9b053485c"
```
tom-saunders-cts: These are currently the same as the english values - presumably we will be sending welsh emails?

src/main/java/uk/gov/hmcts/probate/service/notification/AutomatedNotificationService.java
```diff
@@ -0,0 +1,101 @@
...
+            }
+        } while (keepSearching);
+    }
+
+    private void processCases(List<CaseDetails> cases, NotificationStrategy strategy,
+                              SecurityDTO securityDTO, List<Long> failedCases) {
+        for (CaseDetails caseDetails : cases) {
+            try {
+                Document sentEmail = strategy.sendEmail(caseDetails);
+                automatedNotificationCCDService.saveNotification(
```
tom-saunders-cts: this is:
- querying elasticsearch
- getting back results which might be out of date
- modifying them
- starting an event (which gets the not out of date case data)
- using the event to write back the modified case data from elasticsearch

i know we have attempted to implement a duplication of the optimistic locking used within ccd itself by using the lastModified field, but this does feel a little like we're making the problem we have with exposing ourselves to stale data and potentially overwriting updates worse

src/main/java/uk/gov/hmcts/probate/schedule/SendNotificationsTask.java
```diff
@@ -0,0 +1,81 @@
...
+                log.info("Perform Send Second Stop Reminder started");
+                automatedNotificationService.sendNotification(secondStopReminderDate, SECOND_STOP_REMINDER);
+                log.info("Perform Send Second Stop Reminder finished");
+            }
+        } catch (ApiClientException e) {
+            log.error(e.getMessage());
+        } catch (Exception e) {
+            log.error("Error on SendNotificationsTask Scheduler Send Second Stop Reminder task {}", e.getMessage());
+        }
+    }
```
tom-saunders-cts, line 212: is the thinking here that we will add more and more notification handling into this base task? (means we only have one cronjob which runs daily so i get the logic)

if so are we concerned that this might end up taking overly long / having unexpected failures earlier causing later sections not to run? (i guess this is why we're just catching generic `Exception`? could still find ourselves hitting `OutOfMemoryError` and bailing though)

src/main/java/uk/gov/hmcts/probate/security/SecurityUtils.java
```diff
@@ -70,19 +72,57 @@ public class SecurityUtils {
...
+                try {
+                    userId = getUserId(authorisation);
+                } catch (FeignException | NullPointerException e) {
+                    log.error("Failed to resolve user ID from token: {}", e.getMessage());
+                }
+            }
+        }
+
+        if (StringUtils.isBlank(authorisation) || StringUtils.isBlank(userId)) {
+            log.error("Unable to resolve security context: missing auth or userId");
```
tom-saunders-cts: is it worth logging which of these is blank for potential diagnostics? (obviously not logging the value, just whether each is blank)

src/main/java/uk/gov/hmcts/probate/service/NotificationService.java
```diff
@@ -699,6 +703,24 @@ private ApplicationType getApplicationType(uk.gov.hmcts.reform.ccd.client.model.
                         : ApplicationType.SOLICITOR);
     }
 
+    private String getChannelChoice(uk.gov.hmcts.reform.ccd.client.model.CaseDetails caseDetails) {
+        if (caseDetails == null || caseDetails.getData() == null) {
```
tom-saunders-cts: are these conditions actually reasonable (i.e. if we're calling this method and either evaluates true are we in a good state)? 

ditto `getInformatioNeededByPost`

src/main/java/uk/gov/hmcts/probate/service/notification/AutomatedNotificationCCDService.java
```diff
@@ -0,0 +1,112 @@
...
+                                 final String caseId,
+                                 final SecurityDTO securityDTO,
+                                 final Document sentEmail,
+                                 final NotificationType notificationType) {
+        log.info("AutomatedNotificationCCDService buildCaseData for case id: {}", caseId);
+        GrantOfRepresentationData data = buildCaseData(caseDetails.getData(), sentEmail, notificationType);
+        log.info("AutomatedNotificationCCDService saveNotification to Case: {}", caseId);
+        try {
+            ccdClientApi.updateCaseAsCaseworker(GRANT_OF_REPRESENTATION, caseId, caseDetails.getLastModified(), data,
+                    EventId.AUTOMATED_NOTIFICATION, securityDTO, EVENT_DESCRIPTION, EVENT_SUMMARY);
```
tom-saunders-cts: if we do extend this to handle other notifications are these descriptions/summaries going to be correct?

src/main/java/uk/gov/hmcts/probate/service/notification/TemplateService.java
```diff
@@ -97,6 +96,19 @@ private boolean requestInfoByPostForPersonalApplication(String channelChoice, Ap
                 || CHANNEL_CHOICE_BULKSCAN.equalsIgnoreCase(channelChoice)
                 || CHANNEL_CHOICE_PAPERFORM.equalsIgnoreCase(channelChoice));
     }
+
+    public String getStopReminderTemplateId(ApplicationType applicationType, LanguagePreference languagePreference,
```
tom-saunders-cts: this feels awkward - i can't really describe it better than that but trying to work out what the end result is is not simple

src/main/java/uk/gov/hmcts/probate/service/notification/NotificationStrategy.java
```diff
@@ -0,0 +1,19 @@
...
+
+public interface NotificationStrategy {
+
+    String getQueryTemplate();
+
+    boolean matchesType(NotificationType type);
+
+    Document sendEmail(CaseDetails caseDetails) throws NotificationClientException;
+
+    boolean isFirstReminder();
```
tom-saunders-cts: if there are other notification types which aren't one of these two then?

src/main/java/uk/gov/hmcts/probate/service/NotificationService.java
```diff
@@ -744,4 +766,37 @@ enum CommonNotificationResult {
...
+        String templateId = templateService.getStopReminderTemplateId(applicationType, languagePreference,
+                getChannelChoice(caseDetails), getInformationNeededByPost(caseDetails), isFirstStopReminder);
+        log.info("sendStopReminderEmail applicationType {}, templateId: {}", applicationType, templateId);
+        Map<String, String> personalisation =
+                automatedNotificationPersonalisationService.getPersonalisation(caseDetails, applicationType);
+        log.info("start sendEmail");
+        SendEmailResponse response =
+                notificationClientService.sendEmail(templateId, emailAddress,
+                        personalisation, caseDetails.getId().toString());
+        log.info("First Stop Reminder email reference response: {}", response.getReference());
```
tom-saunders-cts: this log is sent for both first and second stop reminder?

src/main/java/uk/gov/hmcts/probate/service/notification/AutomatedNotificationService.java
```diff
@@ -98,4 +98,8 @@ private void processCases(List<CaseDetails> cases, NotificationStrategy strategy
     private String getLastId(SearchResult searchResult) {
         return searchResult.getCases().getLast().getId().toString();
     }
+
+    private String getErrorMessage(NotificationType type, String date) {
```
tom-saunders-cts: this method is only used in one place? is there a reason to not just have:
```
final String errMsg = String.format("Error...", type, date);
log.error(errMsg, e);
```

tom-saunders-cts (commented):
marking as viewed but expecting further chnages here?

src/main/java/uk/gov/hmcts/probate/service/notification/AutomatedNotificationService.java
FeliTam: updated

src/main/java/uk/gov/hmcts/probate/config/ClockConfiguration.java
```diff
@@ -0,0 +1,15 @@
...
+
+import java.time.Clock;
+
+@Configuration
+public class ClockConfiguration {
+
+    @Bean
+    public Clock clock() {
+        return Clock.systemUTC();
+    }
```
tom-saunders-cts, line 14: this config/bean will likely clash with the Clock bean added as part of DTSPB-2859 / https://github.com/hmcts/probate-back-office/pull/3092

tom-saunders-cts: specifically: https://github.com/hmcts/probate-back-office/pull/3092/files#diff-1825ca99f423e4a1e5b8444b0172250c6854860583013b835c6a2def273b9ff3R14-R17

tom-saunders-cts: That implementation uses the `systemDefaultZone` (which is what `LocalDate.now()` and friends will use) but I'm wondering whether using `systemUTC` like this might be better from a consistency perspective

jamiealbertelli (approved):

PR #3164, opened 2025-06-03
DO NOT MERGE: Playwright: typescript migration

JayAshton:
Jira link

DTSPB-4823

- Swaps playwright to use typescript
- Changes to linting mean that only the playwright directory is being linted & only typescript files
- Adds common playwright lib
- Removed hardcoded credentials from both codecept and playwright tests


FeliTam:
covered by another PR

PR #3193, opened 2025-06-17
DTSPB-4710 Declaration not signed notification

FeliTam:
[DTSPB-4710](https://tools.hmcts.net/jira/browse/DTSPB-4710)

### Change description ###
New automated notification: Declaration not signed

```
[ ] Yes
[x] No
```


src/test/java/uk/gov/hmcts/probate/schedule/SendNotificationsTaskTest.java
```diff
@@ -98,18 +115,21 @@ void setUp() {
                 hseReminderNotification,
                 dormantWarningNotification,
                 unsubmittedApplicationNotification,
+                declarationNotSignedNotification,
                 56,
                 28,
                 30,
                 150,
                 28,
+                3,
```
tom-saunders-cts: i wonder whether having these values as named constants (which we could then use at line 99 for example) might be clearer

src/main/java/uk/gov/hmcts/probate/service/NotificationService.java
```diff
@@ -867,4 +887,87 @@ public void sendUnsubmittedApplicationEmail(uk.gov.hmcts.reform.ccd.client.model
...
+        log.info("Preparing to send declarationNotSigned email to executors for case id: {}", caseId);
+        String templateId = templateService.getDeclarationNotSignedTemplateId(languagePreference, false);
+        for (CollectionMember<ExecutorApplying> executorApplying : unsignedExecutorList) {
+            try {
+                String emailAddress = executorApplying.getValue().getApplyingExecutorEmail();
+                personalisation
+                        .put(PERSONALISATION_EXECUTOR_NAME, executorApplying.getValue().getApplyingExecutorName());
+                sendEmail(emailAddress, templateId, personalisation, caseId);
+            } catch (NotificationClientException e) {
+                log.error("Failed to send declarationNotSigned email to executor: {} for case id: {}",
```
tom-saunders-cts: the exec name is PII - probably should log the executor ID / hashed email instead

tom-saunders-cts (changes requested):
only thing which flagged for me was the PII logging - i thik otherwise this is fine but i've not gone through with a fine-toothed comb

src/test/java/uk/gov/hmcts/probate/schedule/SendNotificationsTaskTest.java
FeliTam: updated

src/main/java/uk/gov/hmcts/probate/service/NotificationService.java
FeliTam: removed exec name and added hashed email

src/main/java/uk/gov/hmcts/probate/service/NotificationService.java
```diff
@@ -962,12 +962,12 @@ private boolean isUnsignedExecutor(CollectionMember<ExecutorApplying> executorMe
                 && !Boolean.TRUE.equals(executor.getApplyingExecutorAgreed());
     }
 
-    private String getExecutorsNamesListString(List<CollectionMember<ExecutorApplying>> executors) {
+    private String[] getExecutorsNamesListString(List<CollectionMember<ExecutorApplying>> executors) {
```
tom-saunders-cts: this is a nit so feel free to ignore, but if we were actually creating an object for serialization i think(?) we would probably prefer a `List<String>` over a `String[]`

tom-saunders-cts (approved):

PR #3269, opened 2025-07-23
DTSPB-4302 Update Print content

MarcusHMCTS:

DTSPB-4302 Update Print content

### Change description ###

```
[ ] Yes
[x ] No
```


src/main/resources/templates/printService/caseDetailsSOL.html
```diff
@@ -277,6 +325,76 @@ <h1>Application for probate for {{ names(case_data.deceasedForenames,case_data.d
     </tr>{% endfor %}
     {% endif %}
 
+    {% if case_data.executorsApplying | length >= 1 %}
```
jamiealbertelli, line 328: Implicitly redundant?

src/main/resources/templates/printService/caseDetailsPA.html
```diff
@@ -40,6 +40,18 @@ <h1>Application for probate for {{ names(case_data.deceasedForenames,case_data.d
     {% endif %}
   </tr>
   <tr>
+  <tr>
+    <td colspan="2">Deceased any other name on will:</td>
+    <td>{% if case_data.deceasedAnyOtherNameOnWill %}{{ case_data.deceasedAnyOtherNameOnWill }}{% else %}N/A{% endif %}</td>
```
tom-saunders-cts: from conversation: might be worth adding a couple of macros `valueOrNA` and `valueOrBlank` - would mean we don't have to check whether there is an else condition to know if the expected outcome is N/A, blank or something else if there isn't a value

src/main/resources/templates/printService/caseDetailsSOL.html
```diff
@@ -148,12 +178,20 @@ <h1>Application for probate for {{ names(case_data.deceasedForenames,case_data.d
     </tr>
     <tr>
       <td colspan="2">Names used by the deceased for other assets:</td>
-      <td>{% for alias in case_data.solsDeceasedAliasNamesList %}{{ alias.value.SolsAliasname }}<br/>{% endfor %}</td>
+        <td>{% if case_data.solsDeceasedAliasNamesList | length == 0 %}N/A</td>
+        {% else %}
+      <td>{% for alias in case_data.solsDeceasedAliasNamesList %}{{ alias.value.SolsAliasname }}<br/>{% endfor %}{% endif %}</td>
```
tom-saunders-cts: the conditional handling here is not correct afaict - the whole condition should be inside the `<td>...</td>` - currently `true` case has two `<td>` tags and `false` has two `</td>` tags

tom-saunders-cts: ditto for below

src/main/resources/templates/printService/caseDetailsSOL.html
```diff
@@ -164,12 +202,22 @@ <h1>Application for probate for {{ names(case_data.deceasedForenames,case_data.d
...
     <tr>
       <td colspan="2">Deceased was domiciled in England or Wales:</td>
       <td>{% if case_data.deceasedDomicileInEngWales %}{{ case_data.deceasedDomicileInEngWales }}{% endif %}</td>
     </tr>
+    {% if case_data.domicilityCountry %}
+    <tr>
+      <td colspan="2">Deceased Domicility Country:</td>
+      <td>{{ case_data.domicilityCountry }}</td>
+    </tr>
+    {% endif %}
```
tom-saunders-cts: from conversation in snowmen i'm wondering if we can find ourselves in odd situations where:
- set `deceasedDomicileInEngWales` = No 
- set `domicilityCountry` = France (e.g.)
- sets `deceasedDomicileInEngWales` = Yes
in that situation should we display this value?

tom-saunders-cts (commented):
comments inline

src/main/resources/templates/printService/caseDetailsSOL.html
```diff
@@ -2,7 +2,10 @@
 <html lang="en">
 <meta charset="UTF-8"/>
 {% macro names(forenames, surname) %}{{ forenames }} {{ surname }}{% endmacro %}
-{% macro address(address)%}
+{% macro address(address) %}
+{% macro valueOrNA(value) %}
+    {% if value %} {{ value }}{% else %}N/A{% endif %}
+{% endmacro %}
```
tom-saunders-cts: these need to be a line higher (it's currently inside the `address` macro)

might be worth adding blank lines separating our multiline macros (and checking whether we can make `names(...)` multilne without breaking the formatting of the document? (possibly need https://pebbletemplates.io/wiki/guide/basic-usage/#whitespace but i'm not sure this is using pebble)

src/main/resources/templates/printService/caseDetailsPA.html
```diff
@@ -87,7 +87,13 @@ <h1>Application for probate for {{ names(case_data.deceasedForenames,case_data.d
     <td colspan="2">Number of executors:</td>
     <td>{{( case_data.executorsApplying | length ) + ( case_data.executorsNotApplying | length ) + 1 }}</td>
   </tr>
-
+  <tr>
+    <td colspan="2">Additional Executors:</td>
+      <td>{% for executor in case_data.solsAdditionalExecutorList %}
+            {{ executor.value.SolsAdditionalExecutorName }}<br/>
+          {% endfor %}
```
tom-saunders-cts, line 95: it's only just occurred to me but should these be using `<ul>...</ul>` rather than `...<br />` ? 

happy to leave this as is unless it's causing problems but i wonder what (e.g.) a screen reader does with this

src/main/resources/templates/printService/caseDetailsSOL.html
```diff
@@ -112,56 +115,74 @@ <h1>Application for probate for {{ names(case_data.deceasedForenames,case_data.d
       <td colspan="2">Solicitor firm name:</td>
       <td>{% if case_data.solsSolicitorFirmName %}{{ case_data.solsSolicitorFirmName }}{% endif %}</td>
     </tr>
-    {% if case_data.solsSolicitorAddress %}
     <tr>
       <td colspan="2">Solicitor firm address:</td>
-      <td>{{address(case_data.solsSolicitorAddress)}}</td>
+      <td>{{ address(case_data.solsSolicitorAddress) }}</td>
```
tom-saunders-cts: without a conditional this will print a blank address block if there is no value - intentional here? (it seems like we should be able to rely on there being an address present but is that always true?)

src/main/resources/templates/printService/caseDetailsSOL.html
```diff
@@ -112,56 +115,74 @@ <h1>Application for probate for {{ names(case_data.deceasedForenames,case_data.d
...
     <tr>
       <td colspan="2">Solicitor Surname:</td>
-      <td>{% if case_data.solsSOTSurname %}{{ case_data.solsSOTSurname }}{% else %}N/A{% endif %}</td>
+      <td>{{ valueOrNA(case_data.solsSOTSurname) }}</td>
     </tr>
+    {% if case_data.solsExecutorAliasNames %}
     <tr>
       <td colspan="2">Solicitor Alias Names:</td>
-      <td>{% if case_data.solsExecutorAliasNames %}{{ case_data.solsExecutorAliasNames }}{% else %}N/A{% endif %}</td>
+      <td>{{ valueOrNA(case_data.solsExecutorAliasNames) }}</td>
```
tom-saunders-cts, line 160: is solsExecutorAliasNames not a collection type of some kind? or have we flattened it into a single string in the case data at this point?

src/main/resources/templates/printService/caseDetailsSOL.html
```diff
@@ -112,56 +115,74 @@ <h1>Application for probate for {{ names(case_data.deceasedForenames,case_data.d
...
       <td colspan="2">Solicitor Is Applying:</td>
-      <td>{% if case_data.solsSolicitorIsApplying %}{{ case_data.solsSolicitorIsApplying }}{% else %}N/A{% endif %}</td>
+      <td>{{ case_data.solsSolicitorIsApplying }}</td>
     </tr>
+    {% endif %}
+    {% if case_data.solsSolicitorNotApplyingReason %}
     <tr>
       <td colspan="2">Solicitor Not Applying Reason:</td>
-      <td>{% if case_data.solsSolicitorNotApplyingReason %}{{ case_data.solsSolicitorNotApplyingReason }}{% else %}N/A{% endif %}</td>
+      <td>{{ valueOrNA(case_data.solsSolicitorNotApplyingReason) }}</td>
```
tom-saunders-cts, line 168: if this doesn't have a value then the condition at l162 will be false and this won't render at all?

src/main/resources/templates/printService/caseDetailsSOL.html
```diff
@@ -112,56 +115,74 @@ <h1>Application for probate for {{ names(case_data.deceasedForenames,case_data.d
...
+      <td>{{ valueOrNA(case_data.solsSOTSurname) }}</td>
     </tr>
+    {% if case_data.solsExecutorAliasNames %}
     <tr>
       <td colspan="2">Solicitor Alias Names:</td>
-      <td>{% if case_data.solsExecutorAliasNames %}{{ case_data.solsExecutorAliasNames }}{% else %}N/A{% endif %}</td>
+      <td>{{ valueOrNA(case_data.solsExecutorAliasNames) }}</td>
     </tr>
+    {% endif %}
+    {% if case_data.solsSolicitorIsApplying %}
```
tom-saunders-cts: is there a reason we are filtering some of these to not show the question and others to show `N/A` when they don't have a value?

src/main/resources/templates/printService/caseDetailsSOL.html
tom-saunders-cts: ditto for `otherExecutorExists` below?

src/main/resources/templates/printService/caseDetailsSOL.html
```diff
@@ -112,56 +115,74 @@ <h1>Application for probate for {{ names(case_data.deceasedForenames,case_data.d
...
+        <td colspan="2">Solicitor Forenames:</td>
+        <td>{{ case_data.solsForenames }}</td>
     </tr>
+    {% endif %}
+    {% if case_data.solsSurname %}
+    <tr>
+        <td colspan="2">Solicitor Surnames:</td>
+        <td>{{ case_data.solsSurname }}</td>
+    </tr>
+    {% endif %}
```
tom-saunders-cts: this has been one of the causes for callbacks to throw exceptions (which we handle a little better than before but still causes problems: https://tools.hmcts.net/jira/browse/DTSPB-4632) - so identifying the fact that a field that several things assume must always have a value is empty might be useful from that perspective.

src/main/resources/templates/printService/caseDetailsSOL.html
```diff
@@ -200,10 +225,12 @@ <h1>Application for probate for {{ names(case_data.deceasedForenames,case_data.d
       <td colspan="2">Deceased date of death:</td>
       <td>{% if case_data.deceasedDateOfDeath %}{{ case_data.deceasedDateOfDeath|date("DD/MM/YYYY", existingFormat="yyyy-MM-dd") }}{% endif %}</td>
     </tr>
+    {% if case_data.dateOfDeathType %}
```
tom-saunders-cts: again, is there a reason we sometimes use N/A and sometimes skip the question?

tom-saunders-cts (changes requested):
one main change (move the `valueOrNA` macro out of the `address` macro definition) and a few comments

src/main/resources/templates/printService/caseDetailsSOL.html
jamiealbertelli: Skipping the question is required so that an empty header doesn't appear in situations where only a PP/CW is authorised and the other is not.

Some fields are only for PP and others CW. N/A is used for both, happy to add N/A to the role specific fields as well.

src/main/resources/templates/printService/caseDetailsSOL.html
tom-saunders-cts: as mentioned elsewhere this is usually where solicitors do not have access but caseworkers do. i'm concerned that at times this might be hiding problems where values are not populated when they should be but this isn't really the place to go through that.

src/main/resources/templates/printService/caseDetailsSOL.html
jamiealbertelli: Change made to display null here to not print the address block in template.

src/main/resources/templates/printService/caseDetailsSOL.html
jamiealbertelli: Fixed the macro here. Will look into blank lines when preview is up.

src/main/resources/templates/printService/caseDetailsPA.html
```diff
@@ -89,10 +90,9 @@ <h1>Application for probate for {{ names(case_data.deceasedForenames,case_data.d
   </tr>
   <tr>
     <td colspan="2">Additional Executors:</td>
-      <td>{% for executor in case_data.solsAdditionalExecutorList %}
-            {{ executor.value.SolsAdditionalExecutorName }}<br/>
-          {% endfor %}
-      </td>
+      <ul>{% for executor in case_data.solsAdditionalExecutorList %}
+          <li>{{ executor.value.SolsAdditionalExecutorName }}</li>
+          {% endfor %}</ul>
```
tom-saunders-cts: from slack - i think we either leave this as it was for minimal change, or we need to keep the `<td>` containing this list.

the original question was mostly from a semantic html perspective - the `<ul>` provides more contextual information (e.g. for a screen reader) but if the ultimate goal for this page is to be printed on paper maybe that's a lower concern than it'd normally be.

src/main/resources/templates/printService/caseDetailsSOL.html
```diff
@@ -39,7 +37,13 @@
...
 {% endmacro %}
 {% macro notApplyingReason(reason) %}
-<td valign="top">{% if reason == "DiedBefore" %}They died before the deceased.{% endif %}{% if reason == "DiedAfter" %}They died after the deceased.{% endif %}{% if reason == "PowerReserved" %}They're holding power reserved.{% endif %}{% if reason == "Renunciation" %}They have renounced.{% endif %}</td>
+<td valign="top">{% if reason == "DiedBefore" %}They died before the deceased.{% endif %}
+    {% if reason == "DiedAfter" %}They died after the deceased.{% endif %}
+    {% if reason == "PowerReserved" %}They're holding power reserved.{% endif %}
+    {% if reason == "Renunciation" %}They have renounced.{% endif %}
+    {% if reason == "PowerOfAttorney" %}They have appointed or wish to appoint another person to act as their attorney.{% endif %}
+    {% if reason == "MentallyIncapable" %}They lack capacity to act as executor and the will allows for a substitution.{% endif %}
+</td>
```
tom-saunders-cts: we spoke briefly about this - i think that changing it is probably good for readability but i wonder if this should be (so we don't find ourselves with _nothing_ if there's no handler):
```suggestion
<td valign="top">{% if reason == "DiedBefore" %}They died before the deceased.
    {% elseif reason == "DiedAfter" %}They died after the deceased.
    {% elseif reason == "PowerReserved" %}They're holding power reserved.
    {% elseif reason == "Renunciation" %}They have renounced.
    {% elseif reason == "PowerOfAttorney" %}They have appointed or wish to appoint another person to act as their attorney.
    {% elseif reason == "MentallyIncapable" %}They lack capacity to act as executor and the will allows for a substitution.
    {% else %}NO TEXT FOR REASON {{ reason }}
    {% endif %}
</td>
```

this also ignores the fact that we probably shouldn't be doing this conversion in this template anyway but that's a different problem

src/main/resources/templates/printService/caseDetailsSOL.html
```diff
@@ -117,7 +121,7 @@ <h1>Application for probate for {{ names(case_data.deceasedForenames,case_data.d
     </tr>
     <tr>
       <td colspan="2">Solicitor firm address:</td>
-      <td>{{ address(case_data.solsSolicitorAddress) }}</td>
+      <td>{% if case_data.solsSolicitorAddress %}{{ address(case_data.solsSolicitorAddress) }}{% else %}N/A{% endif %}</td>
```
tom-saunders-cts, line 128: maybe worth sticking a comment in to say we can't use `valueOrNA(address(x))` here because `address(x)` will return something even if `x` is null or unset

tom-saunders-cts (commented):
comments inline

src/main/resources/templates/printService/caseDetailsSOL.html
jamiealbertelli: Looking into the schema and case data, it's a single string.

src/main/resources/templates/printService/caseDetailsSOL.html
```diff
@@ -37,12 +37,15 @@
...
-    {% if reason == "PowerOfAttorney" %}They have appointed or wish to appoint another person to act as their attorney.{% endif %}
-    {% if reason == "MentallyIncapable" %}They lack capacity to act as executor and the will allows for a substitution.{% endif %}
+<td valign="top">
+    {% if reason == "DiedBefore" %}They died before the deceased.
+        {% elseif reason == "DiedAfter" %}They died after the deceased.
+        {% elseif reason == "PowerReserved" %}They're holding power reserved.
+        {% elseif reason == "Renunciation" %}They have renounced.
+        {% elseif reason == "PowerOfAttorney" %}They have appointed or wish to appoint another person to act as their attorney.
+        {% elseif reason == "MentallyIncapable" %}They lack capacity to act as executor and the will allows for a substitution.
+        {% else %}NO TEXT FOR REASON {{ reason }}
```
tom-saunders-cts: i realize this is literally what i suggested but perhaps we can make this a little more user friendly :)

src/main/resources/templates/printService/caseDetailsSOL.html
```diff
@@ -360,73 +348,50 @@ <h1>Application for probate for {{ names(case_data.deceasedForenames,case_data.d
...
         </tr>
-      {% endif %}
-      {% if case_data.primaryApplicantRelationshipToDeceased %}
         <tr>
           <td colspan="2">Primary Applicant Relationship to Deceased</td>
           <td>{{ case_data.primaryApplicantRelationshipToDeceased }}</td>
-        </tr>
-      {% endif %}
-      {% if case_data.paRelationshipToDeceasedOther %}
+        </tr>}
```
tom-saunders-cts: i'm guessing we don't want the `}` here?

tom-saunders-cts (changes requested):
minor typo and a text change - i'm not sure exactly what that should be but probably a more normally formatted message ight be preferable?

src/main/resources/templates/printService/caseDetailsSOL.html
tom-saunders-cts: "No formatted reason available for value: {{ reason }}." maybe

tom-saunders-cts (approved):

PR #3281, opened 2025-07-28
DTSPB-4726 Add title and clearing type Firm ceased trading with no successor firm

FeliTam:
DTSPB-4726

### Change description ###
Add title and clearing type Firm ceased trading with no successor firm

```
[ ] Yes
[x] No
```


src/functionalTest/java/uk/gov/hmcts/probate/functional/documents/LegalStatementGenerationTests.java
```diff
@@ -105,154 +108,174 @@ public class LegalStatementGenerationTests extends DocumentGenerationTestBase {
...
-
-        assertTrue(!response.contains(DECLARATION_CRIMINAL_WORDING_SINGLE_EXEC));
+        assertAll(
+                () -> assertTrue(response.contains(LEGAL_STATEMENT)),
+                () -> assertTrue(response.contains(DECLARATION_CIVIL_WORDING)),
+                () -> assertTrue(!response.contains(AUTHORISED_SOLICITOR)),
+                () -> assertTrue(response.contains(LEGAL_STATEMENT_DIED_ON)),
+                () -> assertTrue(response.contains(LEGAL_STATEMENT_GOP)),
+                () -> assertTrue(response.contains(PRIMARY_APPLICANT_STATEMENT)),
+                () -> assertTrue(!response.contains(DECLARATION_CRIMINAL_WORDING_SINGLE_EXEC))
```
tom-saunders-cts: i realize it's a pain to change, but if we use Hamcrest's `assertThat(response, containsString(LEGAL_STATEMENT))` etc. we get error messages which tell us _what_ has failed rather than just that `false != true`.

Alternatively `assertTrue(response.contains(LEGAL_STATEMENT), "response does not contain LEGAL_STATEMENT")` would at least indicate which of the checks failed and is arguably a little less verbose (by default the first will report the full response and the value it didn't include)

tom-saunders-cts: there's a different point here that these tests are not really the ideal way to test these behaviours, but that is a separate discussion

tom-saunders-cts (commented):
i think the changes here are reasonable to achieve what has been asked - i'm not entirely certain there are no unexpected changes in our legal statement - the isolated changes seem fine but that template is a nightmare (as Jamie has discovered)

src/functionalTest/java/uk/gov/hmcts/probate/functional/documents/LegalStatementGenerationTests.java
FeliTam: Thanks Tom, updated to  use assertThat for improved error messages

tom-saunders-cts (commented):
again - seems reasonable. are we waiting for the other legal statement changes to land first, or are we thinking we do both in this one change?

MarianIrudayanathan (approved):

PR #3305, opened 2025-08-07
DTSPB-4856 Add Aliases to Lev Event

MarcusHMCTS:

[DTSPB-4856](https://tools.hmcts.net/jira/browse/DTSPB-4856)

### Change description ###

Add Aliases to Lev Event

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/probate/service/DeathRecordCCDService.java
```diff
@@ -50,9 +54,35 @@ public DeathRecord mapDeathRecord(V1Death v1Death) {
...
+                .dateOfDeath(deceased.getDateOfDeath())
+                .aliases(getAliases(deceased));
         }
 
         return builder.build();
     }
-}
+
+    @NotNull
+    static List<uk.gov.hmcts.reform.probate.model.cases.CollectionMember<Alias>> getAliases(Deceased deceased) {
```
tom-saunders-cts: is there a reason to make this static? doing so makes it somewhat more annoying to mock if we need

if this is really only used in DeathRecordService, why not put it there? (edit - it's obviously used 8 lines above...)

tom-saunders-cts: perhaps make this a method `mapAliases` (similar to the other methods here)?

tom-saunders-cts: @FeliTam  i think your solution is better than the above

tom-saunders-cts (commented):
i've really only looked at the direct functional change - the approach seems reasonable - assuming there's not anything weird lurking in the test handling this seems good

FeliTam (approved):

PR #3338, opened 2025-08-21
DTSPB-4893 NFI data share

FeliTam:
DTSPB-4893

### Change description ###
New scheduler job for NFI data share

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/probate/controller/DataExtractControllerIT.java
+++ b/src/integrationTest/java/uk/gov/hmcts/probate/controller/DataExtractControllerIT.java
@@ -41,7 +41,6 @@ import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.
 class DataExtractControllerIT {
 
     public static final String PERFORM_HMRC_DATA_EXTRACT_FINISHED = "Perform HMRC data extract finished";
-    public static final String PERFORM_IM_DATA_EXTRACT_FINISHED = "Perform Iron Mountain data extract finished";
 
     @MockitoBean
     private HmrcDataExtractService hmrcDataExtractService;
@@ -183,5 +182,17 @@ class DataExtractControllerIT {
     void smeeAndFordShouldReturnErrorWithNoDateOnPathParam() throws Exception {
         mockMvc.perform(post("/data-extract/smee-and-ford"))
             .andExpect(status().is4xxClientError());
+    }
+
+    @Test
+    void nfiShouldReturnOkResponseOnValidDateRangeFormat() throws Exception {
+        mockMvc.perform(post("/data-extract/nfi?fromDate=2019-02-13&toDate=2019-02-13"))
+                .andExpect(status().isAccepted());
+    }
+
+    @Test
+    void nfiShouldReturnErrorWithNoDateOnPathParam() throws Exception {
+        mockMvc.perform(post("/data-extract/nfi"))
+                .andExpect(status().is4xxClientError());
     }
 }
--- a/src/integrationTest/java/uk/gov/hmcts/probate/controller/DataExtractControllerIT.java
+++ b/src/integrationTest/java/uk/gov/hmcts/probate/controller/DataExtractControllerIT.java
@@ -44,7 +44,7 @@ import static uk.gov.hmcts.probate.model.DataExtractType.NATIONAL_FRAUD_INITIATI
 @AutoConfigureMockMvc
 class DataExtractControllerIT {
 
-    public static final String PERFORM_HMRC_DATA_EXTRACT_FINISHED = "Perform HMRC data extract called";
+    public static final String PERFORM_HMRC_DATA_EXTRACT_FINISHED = "Perform HMRC data extract task submitted";
 
     @MockitoBean
     private HmrcDataExtractService hmrcDataExtractService;
@@ -72,7 +72,7 @@ class DataExtractControllerIT {
     void ironMountainShouldReturnOkResponseOnValidDateFormat() throws Exception {
         mockMvc.perform(post("/data-extract/iron-mountain?date=2019-03-13"))
             .andExpect(status().isAccepted())
-            .andExpect(content().string("Perform Iron Mountain data extract called"));
+            .andExpect(content().string("Perform Iron Mountain data extract task submitted"));
     }
 
     @Test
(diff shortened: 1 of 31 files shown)


src/integrationTest/java/uk/gov/hmcts/probate/controller/DataExtractControllerIT.java
```diff
@@ -184,4 +183,16 @@ void smeeAndFordShouldReturnErrorWithNoDateOnPathParam() throws Exception {
...
+    void nfiShouldReturnOkResponseOnValidDateRangeFormat() throws Exception {
+        mockMvc.perform(post("/data-extract/nfi?fromDate=2019-02-13&toDate=2019-02-13"))
+                .andExpect(status().isAccepted());
+    }
+
+    @Test
+    void nfiShouldReturnErrorWithNoDateOnPathParam() throws Exception {
+        mockMvc.perform(post("/data-extract/nfi"))
+                .andExpect(status().is4xxClientError());
+    }
```
tom-saunders-cts, line 208: this feels like it's covering the barebones of the system here - we're not actually checking that we integrate with any of the other components meaningfully (i.e. we go and query elasticsearch, we upload to the storage service) or similar here

src/main/java/uk/gov/hmcts/probate/blob/component/BlobUpload.java
```diff
@@ -16,36 +16,35 @@
...
 
-    @Value(value = "${blobstorage.connection}")
-    private String storageConnectionString;
-
-    //Create a unique name for the container
-    String containerName = "smee-and-ford-document-feed";
+    @Value(value = "${blobstorage.connection.smeeandford}")
+    private String smeeAndFordStorageConnectionString;
 
     public void uploadFile(File blobFile) {
```
tom-saunders-cts: this feels like it'd be safer to remove rather than provide a potentially unexpected default?

src/main/java/uk/gov/hmcts/probate/controller/DataExtractController.java
```diff
@@ -124,4 +127,22 @@ public ResponseEntity initiateSmeeAndFordExtract(
...
+            @RequestParam(value = "toDate") String toDate) {
+
+        dataExtractDateValidator.dateValidator(fromDate, toDate);
+
+        log.info("Calling perform NFI data extract from date...");
+        ExecutorService executor = Executors.newFixedThreadPool(1);
+        executor.submit(() -> {
+            dataExtractService.performExtractForDateRange(fromDate, toDate, NATIONAL_FRAUD_INITIATIVE);
+        });
+        log.info("Perform NFI data extract from date finished");
```
tom-saunders-cts: strictly this may not have finished - we've only submitted it to the executor and it could fail

tom-saunders-cts: do we not already have a threadpool for work like this? it seems odd to created one with one thread which is only used for one job

src/main/java/uk/gov/hmcts/probate/service/zip/ZipFileService.java
```diff
@@ -73,6 +74,22 @@ public class ZipFileService {
         WELSH_INTESTACY_GRANT_REISSUE, WELSH_ADMON_WILL_GRANT_REISSUE, WELSH_AD_COLLIGENDA_BONA_GRANT_REISSUE};
     private static final String HEADER_ROW_FILE = "templates/dataExtracts/ManifestFileHeaderRow.csv";
 
+    public File generateZipFile(List<ReturnedCaseDetails> cases, File tempFile, String date, DataExtractType type) {
```
tom-saunders-cts, line 79: having two methods with identical names and only a single parameter difference but which do completely different things is a way to make someone have a bad day in the future i suspect

src/main/java/uk/gov/hmcts/probate/service/dataextract/DataExtractService.java
```diff
@@ -0,0 +1,54 @@
...
+            List<ReturnedCaseDetails> cases =
+                    caseQueryService.findAllCasesWithGrantIssuedDate(String.valueOf(type), date);
+            log.info("ES query executed for date: {}, cases found: {}", date, cases.size());
+            File zipFile = strategy.generateZipFile(cases, date);
+            log.info("Zip file generated for type: {} date: {} filename: {}", type, date, zipFile.getName());
+            strategy.uploadToBlobStorage(zipFile);
+            log.info("Data extract for type: {} date: {} completed successfully", type, date);
+            log.info("Zipfile exists: {}", zipFile.exists());
+        } catch (Exception e) {
+            final String errorMsg = String.format("Error running data extract for type: %s for date: %s", type, date);
```
tom-saunders-cts, line 49: are we expecting to add alerting or similar to identify when this doesn't work?

tom-saunders-cts (commented):
comments inline - overall seems reasonable approach, but are we setting ourselves up for future pain with the leftover handling for uploads?

src/main/java/uk/gov/hmcts/probate/service/dataextract/SmeeAndFOrdDataExtractStrategy.java
```diff
@@ -0,0 +1,52 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.List;
+
+import static uk.gov.hmcts.probate.model.DataExtractType.SMEE_AND_FORD;
+
+@Slf4j
+@Service
+@RequiredArgsConstructor
+public class SmeeAndFOrdDataExtractStrategy implements  DataExtractStrategy {
```
tom-saunders-cts: FOrd -> Ford ?

FeliTam: fixed typo

src/main/java/uk/gov/hmcts/probate/controller/DataExtractController.java
FeliTam: Added AsyncConfig to manage the threadpool. Updated logging.

src/main/java/uk/gov/hmcts/probate/blob/component/BlobUpload.java
FeliTam: Moved container and connectionString to SmeeAndFordDataExtractStrategy.
Other methods not fully migrated, and will leave it to future enhancement.

src/test/java/uk/gov/hmcts/probate/service/dataextract/SmeeAndFordExtractStrategyTest.java
```diff
@@ -0,0 +1,62 @@
...
+                () -> assertEquals(SMEE_AND_FORD, strategy.getType())
+        );
+    }
+
+    @Test
+    void uploadToBlobStorageShouldUploadAndDeleteFile() throws Exception {
+        File tempFile = Files.createFile(tempDir.resolve("test-upload.txt")).toFile();
+
+        SmeeAndFordDataExtractStrategy sut = new SmeeAndFordDataExtractStrategy(blobUpload);
+        var field = SmeeAndFordDataExtractStrategy.class.getDeclaredField("smeeAndFordStorageConnectionString");
```
tom-saunders-cts: having to use reflection to access class internals for testing suggests we might be better to have those injected through a constructor?

tom-saunders-cts (dismissed):
generally seems fine - there are a couple of points which might be worth thinking about but nothing critical to my eye

tom-saunders-cts (dismissed):
Looks reasonable as an approach, thanks for the constructor injection changes

MarcusHMCTS (approved):

PR #3339, opened 2025-08-21
Fix Caveat search issue

IswaryaPepakayala:
### Change description ###

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/probate/service/CaveatQueryService.java
```diff
@@ -60,6 +63,9 @@ public class CaveatQueryService {
     private final AuthTokenGenerator serviceAuthTokenGenerator;
     private final SecurityUtils securityUtils;
     private final BusinessValidationMessageRetriever businessValidationMessageRetriever;
+    @Value("${data-extract.pagination.size}")
+    protected int dataExtractPaginationSize;
```
tom-saunders-cts, line 67: would making this a constructor injected parameter make this easier to test? (possibly not since we can just make the mock call only return a couple of responses and total of 10 to ensure we're looping correctly?)

src/main/java/uk/gov/hmcts/probate/service/CaveatQueryService.java
```diff
@@ -92,12 +99,45 @@ public List<ReturnedCaveatDetails> findCaveatDraftCases(String startDate,
...
+                                                             String queryDateStart, String queryDateEnd,
+                                                               CaseType caseType) {
+
+        List<ReturnedCaveatDetails> allResults = new ArrayList<>();
+        List<ReturnedCaveatDetails> pagedResults;
+        int index = 0;
+        int pageStart = 0;
+        int total = 10000000;
+        String paginatedQry = jsonQuery;
+        while (index < total) {
```
tom-saunders-cts, line 119: do we have any test coverage for how this actually functions if multiple requests are made? it seems like it's going to be a messy one to implement since we can't just mock out `runQuery()` at line 122, and looking into what that actually does we would end up needing to mock the `restTemplate.postForObject()` call

tom-saunders-cts (dismissed):

src/main/java/uk/gov/hmcts/probate/service/NotificationService.java
```diff
@@ -745,9 +743,9 @@ private String removedSolicitorNameForPersonalisation(CaseData caseData) {
                 caseData.getRemovedRepresentative().getSolicitorLastName()) : null;
     }
 
-    private String getTemplateId(ApplicationType applicationType) {
+    private String getTemplateId() {
```
tom-saunders-cts, line 746: it's preexisting but maybe this could be named less generically? (and moved out to be a method of `TemplateService`?)

tom-saunders-cts (approved):
minor comment re: naming/location of method but since preexists i'm not going to force the change

PR #3379, opened 2025-09-11
DTSPB-4821 Update grant type notifications

jamiealbertelli:

### Change description ###

```
[ ] Yes
[ ] No
```


src/main/java/uk/gov/hmcts/probate/service/NotificationService.java
```diff
@@ -1043,4 +1058,24 @@ private List<String> getExecutorsNamesList(List<CollectionMember<ExecutorApplyin
...
     }
+
+    private String getCaseTypeText(String caseType, LanguagePreference languagePreference) {
+        if (languagePreference.equals(LanguagePreference.ENGLISH)) {
+            return switch (caseType) {
+                case GOP_CASE_TYPE -> "grant of probate";
+                case INTESTACY_CASE_TYPE -> "letters of administration";
+                case ADMON_WILL_CASE_TYPE -> "letters of administration with will annexed";
+                case AD_COLLIGENDA_BONA_CASE_TYPE -> "Ad Colligenda Bona grant";
+                default -> "grant of probate";
```
tom-saunders-cts: are we better to return a potentially incorrect string or log an error, throw, and abort whatever operation is happening?

tom-saunders-cts: also is there some reason we title case Ad Colligenda Bona but not any other grant type?

src/main/java/uk/gov/hmcts/probate/service/NotificationService.java
```diff
@@ -155,6 +159,16 @@ public Document sendEmail(State state, CaseDetails caseDetails, Optional<CaseOri
...
 
+        if (caseData.getApplicationType().equals(ApplicationType.SOLICITOR)
+                && (state == State.GRANT_ISSUED || state == State.GRANT_ISSUED_INTESTACY
+                || state == State.GRANT_REISSUED)) {
+            personalisation.put("case_type_text", getCaseTypeText(caseData.getCaseType(), LanguagePreference.ENGLISH));
+            if (caseData.getLanguagePreference() == LanguagePreference.WELSH) {
+                personalisation.put("welsh_case_type_text",
+                        getCaseTypeText(caseData.getCaseType(), LanguagePreference.WELSH));
+            }
+        }
```
tom-saunders-cts: i generally find our email handling nontrivial to reason about as we're trying to do the right thing for every option in a single path - not to say this is wrong or that this ticket should be used to unpick that, but i do wonder if it's something we might need to consider longer term.

(it also makes actually knowing that we have adequate test coverage to ensure that we're not missing parameters etc. for each template more difficult.)

src/integrationTest/java/uk/gov/hmcts/probate/service/NotificationServiceIT.java
```diff
@@ -2578,6 +2581,182 @@ void sendDeclarationNotSignedEmail() throws NotificationClientException {
...
+                .build(), null, CASE_ID);
+
+        when(notificationClient.sendEmail(any(), any(), any(), any())).thenReturn(sendEmailResponse);
+
+        notificationService.sendEmail(State.GRANT_ISSUED_INTESTACY, caseDetails);
+
+        ArgumentCaptor<Map<String, Object>> personalisationCaptor = ArgumentCaptor.forClass(Map.class);
+        verify(notificationClient).sendEmail(any(), any(), personalisationCaptor.capture(), any());
+
+        Map<String, Object> personalisation = personalisationCaptor.getValue();
```
tom-saunders-cts: it's a pedantic point but we should probably check that:
a) the template we're passing is the one we expect
b) we have all the parameters the template is expecting (`deceased_name`, `solicitor_name`, `solicitor_reference`, `ccd_reference`, `case_type_text`, `registry_name`, `registry_phone`)

in theory we should probably also be checking (though this obviously fails for our current implementation) that:
c) we aren't sending anything else

src/integrationTest/java/uk/gov/hmcts/probate/service/NotificationServiceIT.java
```diff
@@ -2578,6 +2581,182 @@ void sendDeclarationNotSignedEmail() throws NotificationClientException {
                 anyString());
     }
 
+    @Test
+    void shouldSetCaseTypeTextForSolicitorGrantOfProbateIssued() throws NotificationClientException {
```
tom-saunders-cts: is there a reason these tests are within the integration tests rather than the unit tests? certainly for this test we don't actually want an email sending (and we aren't actually testing if it did in any way?)

src/main/java/uk/gov/hmcts/probate/service/NotificationService.java
jamiealbertelli: Will throw an error here. 
Greta confirmed the capitalisation is intentional

src/test/java/uk/gov/hmcts/probate/service/NotificationServiceTest.java
```diff
@@ -401,4 +402,238 @@ void sendStopResponseReceivedEmailThrowsExceptionWhenEmailIsNull() throws Notifi
...
+        when(caseData.getPrimaryApplicantEmailAddress()).thenReturn("primary@probate-test.com");
+        when(caseData.getSolsSolicitorEmail()).thenReturn("abc@gmail.com");
+
+        Map<String, Object> personalisation = new HashMap<>();
+        personalisation.put("case_type_text", "grant of probate");
+
+        notificationService.updatePersonalisationForSolicitorGrantIssuedEmails(State.GRANT_ISSUED, caseData,
+                personalisation);
+
+        assertEquals("grant of probate", personalisation.get("case_type_text"));
```
tom-saunders-cts, line 463: combined with line 419 this is really only testing that updating the personalisation doesn't remove this - is that what we're really checking here?

src/main/java/uk/gov/hmcts/probate/service/NotificationService.java
```diff
@@ -213,6 +219,19 @@ void updatePersonalisationForSolicitor(CaseData caseData, Map<String, Object> pe
         }
     }
 
+    void updatePersonalisationForSolicitorGrantIssuedEmails(State state, CaseData caseData,
```
tom-saunders-cts: should this return the map rather than void? it works either way but i am always a little wary of methods without declared return types

src/main/java/uk/gov/hmcts/probate/service/NotificationService.java
```diff
@@ -1043,4 +1062,34 @@ private List<String> getExecutorsNamesList(List<CollectionMember<ExecutorApplyin
             .map(ExecutorApplying::getApplyingExecutorName)
             .toList();
     }
+
+    private String getCaseTypeText(String caseType, LanguagePreference languagePreference) {
```
tom-saunders-cts: i wonder if combining the two new methods makes sense? we're essentially doing:
```
if (should call getCaseTypeText) {
  map.en = getCaseTypeText(caseType, ENGLISH);
  if (welsh) {
    map.cy = getCaseTypeText(caseType, WELSH);
  }
}
```

and then doing the same `if (welsh)` inside here

could we instead have:
```
if (should call) {
  if (caseType.isBlank()) { throw }
  map.en = switch (caseType) {
    ...
  };
  if (welsh) {
    map.cy = switch (caseType) {
      ..
    }
  }
}

File: src/main/java/uk/gov/hmcts/probate/service/NotificationService.java [hunk 8]
```diff
@@ -156,14 +160,16 @@ public Document sendEmail(State state, CaseDetails caseDetails, Optional<CaseOri
         }
 
         updatePersonalisationForSolicitor(caseData, personalisation);
+        updatePersonalisationForSolicitorGrantIssuedEmails(state, caseData, personalisation);
```
tom-saunders-cts: we should use the return value here - currently the behaviour is that it mutates the parameter directly and then returns the same reference, but if that ever changes we're going to have an issue

src/main/java/uk/gov/hmcts/probate/service/NotificationService.java
```diff
@@ -213,6 +219,45 @@ void updatePersonalisationForSolicitor(CaseData caseData, Map<String, Object> pe
...
 
+    Map<String, Object> updatePersonalisationForSolicitorGrantIssuedEmails(State state, CaseData caseData,
+                                                            Map<String, Object> personalisation) {
+        if (caseData.getApplicationType().equals(ApplicationType.SOLICITOR)
+                && (state == State.GRANT_ISSUED || state == State.GRANT_ISSUED_INTESTACY
+                || state == State.GRANT_REISSUED)) {
+
+            String caseType = caseData.getCaseType();
+            if (caseType.isBlank()) {
+                log.error("Personalisation validation failed for blank caseType");
```
tom-saunders-cts: i would consider adding the case reference (so we can go backwards if needed from the logs to the case) - same for the log entries below

src/test/java/uk/gov/hmcts/probate/service/NotificationServiceTest.java
```diff
@@ -401,4 +402,227 @@ void sendStopResponseReceivedEmailThrowsExceptionWhenEmailIsNull() throws Notifi
...
+        when(caseData.getCaseType()).thenReturn("gop");
+        when(caseData.getApplicationType()).thenReturn(ApplicationType.SOLICITOR);
+        when(caseData.getLanguagePreference()).thenReturn(LanguagePreference.ENGLISH);
+        when(caseData.getSolsSOTName()).thenReturn("Solicitor Name");
+        when(caseData.getPrimaryApplicantEmailAddress()).thenReturn("primary@probate-test.com");
+        when(caseData.getSolsSolicitorEmail()).thenReturn("abc@gmail.com");
+
+        Map<String, Object> personalisation = new HashMap<>();
+
+        notificationService.updatePersonalisationForSolicitorGrantIssuedEmails(State.GRANT_ISSUED, caseData,
```
tom-saunders-cts: ditto comment in implementation - use the return value here rather than the side effect of modifying the input

src/test/java/uk/gov/hmcts/probate/service/NotificationServiceTest.java
```diff
@@ -401,4 +402,227 @@ void sendStopResponseReceivedEmailThrowsExceptionWhenEmailIsNull() throws Notifi
...
+    void updatePersonalisationText_ForGrantIssuedCaseTypeGop() {
+        CaseData caseData = mock(CaseData.class);
+        CaseDetails caseDetails = mock(CaseDetails.class);
+        when(caseDetails.getData()).thenReturn(caseData);
+        when(caseData.getCaseType()).thenReturn("gop");
+        when(caseData.getApplicationType()).thenReturn(ApplicationType.SOLICITOR);
+        when(caseData.getLanguagePreference()).thenReturn(LanguagePreference.ENGLISH);
+        when(caseData.getSolsSOTName()).thenReturn("Solicitor Name");
+        when(caseData.getPrimaryApplicantEmailAddress()).thenReturn("primary@probate-test.com");
+        when(caseData.getSolsSolicitorEmail()).thenReturn("abc@gmail.com");
```
tom-saunders-cts: i think we only actually need:
```suggestion
        CaseData caseData = mock(CaseData.class);
        when(caseData.getCaseType()).thenReturn("gop");
        when(caseData.getLanguagePreference()).thenReturn(LanguagePreference.ENGLISH);
```
here? (and probably below)

tom-saunders-cts (dismissed):
i think we've addressed all the concerns i had here - sorry it's been such a back and forth

tom-saunders-cts (approved):

PR #3382, opened 2025-09-12
DTSPB-4947 Fee Increase 2025

MarcusHMCTS:

[DTSPB-4947](https://tools.hmcts.net/jira/browse/DTSPB-4947)

### Change description ###

```
[ ] Yes
[x ] No
```


tom-saunders-cts (commented):
is this something we can parameterize so we can have this controlled via feature flag or similar? otherwise we're just going ot have to have this change pending and make sure it gets deployed when the fee changes

tom-saunders-cts (dismissed):
seems like this is more complex to make dynamic in the way we can for frontend - also presumably we have better/alternative ways to message this to professional users

tom-saunders-cts (dismissed):
new fee value updated per new information

tom-saunders-cts (dismissed):
further updating the value

tom-saunders-cts (approved):

PR #3405, opened 2025-09-22
Dtspb 4954 draft state with payment report

MarcusHMCTS:

[DTSPB-4954](https://tools.hmcts.net/jira/browse/DTSPB-4954)

### Change description ###

```
[ ] Yes
[x] No
```


src/main/resources/templates/elasticsearch/caseMatching/draft_cases_date_range_query.json
```diff
@@ -24,11 +22,10 @@
...
     {
-      "id": {
-        "order": "asc"
-      }
+      "reference.keyword": "asc"
     }
-  ]
-}
\ No newline at end of file
+  ]
```
tom-saunders-cts, line 31: i know that this is consistent with (and required by) the existing handling, but having a partial template of the required output with the order of template parameters being important feels so, so fragile

src/main/java/uk/gov/hmcts/probate/service/FetchDraftCaseService.java
```diff
@@ -26,76 +29,85 @@ public class FetchDraftCaseService {
     private final ServiceRequestClient serviceRequestClient;
     private final NotificationService notificationService;
     private final CaveatQueryService caveatQueryService;
+    private final ElasticSearchRepository elasticSearchRepository;
+    private static final String DRAFT_CASES_QUERY = "templates/elasticsearch/caseMatching/"
+            + "draft_cases_date_range_query.json";
 
-    public void fetchGORCases(String startDate, String endDate) {
+    public void fetchDraftCases(String startDate, String endDate,boolean isCaveat) {
```
tom-saunders-cts: i would generally avoid having a method whose behaviour is controlled by a boolean - it's almost never clear from the usage context what `true` or `false` means (and even when a parameter name is available it's not always reliable - see `deceasedAnyOtherNameOnWill` which stores `Yes` if the name is the _same_ on the will). 

if you want to change this to take a `CcdCaseType` (and then deal with the fact that this only handles two of the case types) I'd be more happy since the meaning is clearer at the call site.

src/main/java/uk/gov/hmcts/probate/service/NotificationService.java
```diff
@@ -288,25 +287,14 @@ public Document sendCaveatNocEmail(State state, CaveatDetails caveatDetails) thr
         return getSentEmailDocument(state, emailAddress, response);
     }
 
-    public void sendEmailForGORSuccessfulPayment(List<ReturnedCaseDetails> cases, String fromDate, String toDate)
+    public void sendEmailForDraftSuccessfulPayment(List<uk.gov.hmcts.reform.ccd.client.model.CaseDetails> cases,
```
tom-saunders-cts, line 299: as above re: boolean flags on methods

src/main/java/uk/gov/hmcts/probate/service/FetchDraftCaseService.java
```diff
@@ -26,76 +29,85 @@ public class FetchDraftCaseService {
...
+    private void processCases(List<CaseDetails> cases, List<CaseDetails> successfulPaymentCases) {
+        for (CaseDetails caseDetails : cases) {
+            log.info("Draft state case id: {}", caseDetails.getId());
+            boolean isPaymentSuccessful = processPayment(caseDetails.getId().toString());
+
+            if (isPaymentSuccessful) {
+                log.info("Payment status is Success for case id: {}", caseDetails.getId());
+                successfulPaymentCases.add(caseDetails);
+            } else {
+                log.info("Payment status is not Success for case id: {}", caseDetails.getId());
```
tom-saunders-cts: obviously this is behaviour which existed before but do we want to have a log entry for every case which doesn't meet the criteria (we could potentially collect all the cases and then log them once, but that introduces more complexity which is arguably unneeded)

src/main/java/uk/gov/hmcts/probate/service/FetchDraftCaseService.java
```diff
@@ -26,76 +29,85 @@ public class FetchDraftCaseService {
...
+            }
         }
     }
 
-    private void sendGORSuccessfulPaymentNotification(List<ReturnedCaseDetails> successfulPaymentCases,
-                                                      String startDate, String endDate) {
+    private String getLastId(SearchResult searchResult) {
+        return searchResult.getCases().getLast().getId().toString();
```
tom-saunders-cts, line 97: is `SearchResult` outside our control? this seems like it should be a method of that object

src/main/java/uk/gov/hmcts/probate/schedule/FetchDraftCasesWithPaymentTask.java
```diff
@@ -38,9 +38,9 @@ public void run() {
         try {
             dataExtractDateValidator.dateValidator(startDate, endDate);
             log.info("Perform send email for GOR draft cases wth payment  from date started");
-            fetchDraftCaseService.fetchGORCases(startDate, endDate);
+            fetchDraftCaseService.fetchDraftCases(startDate, endDate, false);
```
tom-saunders-cts: what is `false` (or `true` below)?

src/main/java/uk/gov/hmcts/probate/service/notification/GrantOfRepresentationPersonalisationService.java
```diff
@@ -187,22 +174,13 @@ private StringBuilder getExelaBuiltData(List<ReturnedCaseDetails> cases) {
...
-
-    private StringBuilder getCaveatDraftCasesBuiltData(List<ReturnedCaveatDetails> cases) {
+    private StringBuilder getDraftCasesBuiltData(List<uk.gov.hmcts.reform.ccd.client.model.CaseDetails> cases) {
         StringBuilder data = new StringBuilder();
 
-        for (ReturnedCaveatDetails currentCase : cases) {
-            getCaseData(data, currentCase.getId(), currentCase.getData().getDeceasedForenames(),
-                    currentCase.getData().getDeceasedSurname());
+        for (uk.gov.hmcts.reform.ccd.client.model.CaseDetails currentCase : cases) {
+            getCaseData(data, currentCase.getId(),
```
tom-saunders-cts, line 183: it's not changed here, but this method doesn't log the fact that it's catching `Exception`, it just silently swallows it, terminates the current record and then continues - this feels like a good way to get junk data and have no record of the fact later

src/main/java/uk/gov/hmcts/probate/service/notification/GrantOfRepresentationPersonalisationService.java
```diff
@@ -101,27 +100,15 @@ public Map<String, Object> getNocPersonalisation(Long caseId, String solicitorNa
         return personalisation;
     }
 
-    public Map<String, Object> getGORDraftCaseWithPaymentPersonalisation(List<ReturnedCaseDetails> cases,
-                                                                         String fromDate, String toDate) {
+    public Map<String, Object> getDraftCaseWithPaymentPersonalisation(
+            List<uk.gov.hmcts.reform.ccd.client.model.CaseDetails> cases,
+            String fromDate, String toDate, boolean isCaveat) {
```
tom-saunders-cts: again re: boolean parameters

tom-saunders-cts (commented):
i've only looked at the implementation here, comments inline

src/main/java/uk/gov/hmcts/probate/service/FetchDraftCaseService.java
MarcusHMCTS: done, thanks

src/main/java/uk/gov/hmcts/probate/schedule/FetchDraftCasesWithPaymentTask.java
MarcusHMCTS: fixed, thanks

src/main/java/uk/gov/hmcts/probate/service/notification/GrantOfRepresentationPersonalisationService.java
MarcusHMCTS: done, thanks

src/main/java/uk/gov/hmcts/probate/service/NotificationService.java
MarcusHMCTS: fixed, thanks

src/main/java/uk/gov/hmcts/probate/service/FetchDraftCaseService.java
```diff
@@ -26,76 +30,86 @@ public class FetchDraftCaseService {
...
-                boolean isPaymentSuccessful = processPayment(returnedCaseDetails.getId().toString());
-
-                if (isPaymentSuccessful) {
-                    log.info("Payment status is Success for case id: {}", returnedCaseDetails.getId());
-                    successfulPaymentCases.add(returnedCaseDetails);
-                } else {
-                    log.info("Payment status is not Success for case id: {}", returnedCaseDetails.getId());
-                }
+            SecurityDTO securityDTO = securityUtils.getUserBySchedulerTokenAndServiceSecurityDTO();
+            String caseTypeName = ccdCaseType.getName().equals(CAVEAT.getName())
```
tom-saunders-cts: if we care about restricting the use to the two options here (which the current implementation here does not) we could do something like:
```suggestion
            String caseTypeName = ccdCaseType.getName().equals(CAVEAT.getName())
            String caseTypeName = switch (ccdCaseType) {
                case CAVEAT, GRANT_OF_REPRESENTATION -> ccdCaseType.getName();
                default -> throw new IllegalArgumentException(...);
            };
```

or if we don't care: 
```suggestion
            String caseTypeName = ccdCaseType.getName();
```

tom-saunders-cts (commented):
again, not looked in detail at testing

src/main/java/uk/gov/hmcts/probate/service/FetchDraftCaseService.java
```diff
@@ -37,11 +34,9 @@ public class FetchDraftCaseService {
     public void fetchDraftCases(String startDate, String endDate, CcdCaseType ccdCaseType) {
         try {
             SecurityDTO securityDTO = securityUtils.getUserBySchedulerTokenAndServiceSecurityDTO();
-            String caseTypeName = ccdCaseType.getName().equals(CAVEAT.getName())
-                    ? CAVEAT.getName() : GRANT_OF_REPRESENTATION.getName();
-            log.info("Fetch {} draft cases from date {} to {}", caseTypeName, startDate, endDate);
+            log.info("Fetch {} draft cases from date {} to {}", ccdCaseType.getName(), startDate, endDate);
```
tom-saunders-cts: i take it we're just accepting that if we call this with one of the other CcdCaseTypes we get possibly unexpected results?

src/test/java/uk/gov/hmcts/probate/schedule/FetchDraftCasesWithPaymentTaskTest.java
```diff
@@ -41,8 +43,7 @@ class FetchDraftCasesWithPaymentTaskTest {
 
     @BeforeEach
     void setUp() {
-        clock = Clock.systemDefaultZone();
-
+        clock = Clock.system(ZoneId.of("Europe/London"));
```
tom-saunders-cts: If we are going to change this to use a different clock it might be preferable to use a fixed clock (and then adapt the tests such that they are expecting the fixed dates that will return
```suggestion
        clock = Clock.fixed( 
                Instant.parse("2025-10-02T03:00:00Z"),
                ZoneOffset.UTC);
```

this would mean that the `DATE_TODAY` value on line 41 would need to be fixed so as not to reference `LocalDate.now()` and instead would be `"2025-10-02"`

src/test/java/uk/gov/hmcts/probate/service/FetchDraftCaseServiceTest.java
```diff
@@ -124,6 +124,41 @@ void fetchCasesWithUnsuccessfulPayment() throws NotificationClientException {
...
+        when(paymentsResponse.getPayments()).thenReturn(List.of(PaymentDto.builder().status("success").build()));
+        when(serviceRequestClient.retrievePayments(anyString(), anyString(), anyString(), anyString()))
+                .thenReturn(paymentsResponse);
+
+        assertDoesNotThrow(() -> fetchDraftCaseService.fetchDraftCases("2023-01-01",
+                "2023-01-31", GRANT_OF_REPRESENTATION));
+
+        verify(notificationService, times(1))
+                .sendEmailForDraftSuccessfulPayment(anyList(), anyString(), anyString(), any());
+    }
```
tom-saunders-cts, line 163: is it worth verifing that:
- `fetchNextPage(...)` is actually called twice?
-  the call to `sendEmailForDraftSuccessfulPayment(...)` has the expected values in the list?

src/test/java/uk/gov/hmcts/probate/schedule/FetchDraftCasesWithPaymentTaskTest.java
```diff
@@ -80,7 +82,40 @@ void shouldPerformDraftCasesExtractForAdhocDateRange() {
...
+        verify(fetchDraftCaseService).fetchDraftCases(adhocDate, DATE_TODAY,GRANT_OF_REPRESENTATION);
+    }
+
+    @Test
+    void shouldPerformDraftCaveatCasesExtractDateRange() {
+        ResponseEntity<String> responseEntity = ResponseEntity.accepted()
+                .body("Perform hmrc data extract from date finished");
+        fetchDraftCasesWithPaymentTask.run();
+        assertEquals(HttpStatus.ACCEPTED, responseEntity.getStatusCode());
+        assertEquals("Perform hmrc data extract from date finished", responseEntity.getBody());
```
tom-saunders-cts: these are asserting about an entity we just created - and which never seems to be used otherwise

src/test/java/uk/gov/hmcts/probate/schedule/FetchDraftCasesWithPaymentTaskTest.java
```diff
@@ -80,7 +82,40 @@ void shouldPerformDraftCasesExtractForAdhocDateRange() {
...
+
+    @Test
+    void shouldPerformDraftCasesCaveatExtractForAdhocDateRange() {
+        ResponseEntity<String> responseEntity = ResponseEntity.accepted()
+                .body("Perform hmrc data extract from date finished");
+        fetchDraftCasesWithPaymentTask.run();
+        assertEquals(HttpStatus.ACCEPTED, responseEntity.getStatusCode());
+        assertEquals("Perform hmrc data extract from date finished", responseEntity.getBody());
+        verify(dataExtractDateValidator).dateValidator(adhocDate, DATE_TODAY);
+        verify(fetchDraftCaseService).fetchDraftCases(adhocDate, DATE_TODAY,CAVEAT);
```
tom-saunders-cts: aren't these three tests identical? (and i guess the three immediately above these are also the same but with `GRANT_OF_REPRESENTATION` rather than `CAVEAT`)

src/test/java/uk/gov/hmcts/probate/schedule/FetchDraftCasesWithPaymentTaskTest.java
```diff
@@ -80,7 +82,40 @@ void shouldPerformDraftCasesExtractForAdhocDateRange() {
...
+
+    @Test
+    void shouldPerformDraftCaveatCasesExtractDateRange() {
+        ResponseEntity<String> responseEntity = ResponseEntity.accepted()
+                .body("Perform hmrc data extract from date finished");
+        fetchDraftCasesWithPaymentTask.run();
+        assertEquals(HttpStatus.ACCEPTED, responseEntity.getStatusCode());
+        assertEquals("Perform hmrc data extract from date finished", responseEntity.getBody());
+        verify(dataExtractDateValidator).dateValidator(adhocDate, DATE_TODAY);
+        verify(fetchDraftCaseService).fetchDraftCases(adhocDate, DATE_TODAY,CAVEAT);
```
tom-saunders-cts, line 60: it feels like we could put this single line into `shouldPerformDraftCasesExtractDateRange` after the check for the `GRANT_OF_REPRESENTATION` call?

tom-saunders-cts (dismissed):

tom-saunders-cts (dismissed):

tom-saunders-cts (approved):
reapproving as GH reports no overall change since previous

PR #3432, opened 2025-10-08
DTSPB-4885 checking for nulls

justiceia:

https://tools.hmcts.net/jira/browse/DTSPB-4885

### Change description ###
making checks before accessing
removed private so i can test directly

```
[ ] Yes
[ ] No
```


src/test/java/uk/gov/hmcts/probate/service/notification/AutomatedNotificationPersonalisationServiceTest.java
```diff
@@ -153,4 +154,39 @@ void getPersonalisationShouldIncludeAllFields() {
...
+            .caseStopSubReasonDocRequired(null)
+            .build();
+        CollectionMember<StopReason> cm1 =
+                new CollectionMember<>(null, r1);
+        CollectionMember<StopReason> cm2 =
+                new CollectionMember<>(null, r2);
+        CollectionMember<StopReason> cm3 =
+                new CollectionMember<>(null, r3);
+        List<CollectionMember<StopReason>>
+                domainList = List.of(cm1, cm2, cm3);
```
tom-saunders-cts: i don't think we need to be quite so aggressive on line breaks here.
also there's a one-arg constructor for `CollectionMember` which implicitly sets the `id` to `null` so maybe that's preferable
```suggestion
        CollectionMember<StopReason> cm1 = new CollectionMember<>(r1);
        CollectionMember<StopReason> cm2 = new CollectionMember<>(r2);
        CollectionMember<StopReason> cm3 = new CollectionMember<>(r3);
        List<CollectionMember<StopReason>> domainList = List.of(cm1, cm2, cm3);
```

tom-saunders-cts: i'm also wondering if we should verify that if there are non-null values after the null it still includes - we don't currently have anything which gets included but if we just stuck a new value at start and end then that might highlight that if there are multiple stop reasons for the same document they get included multiple times

justiceia: added changes

tom-saunders-cts (dismissed):

tom-saunders-cts (approved):

PR #3440, opened 2025-10-10
- Updated ComplexTypes.json to change field types for solicitor email…

NewmanJustice:
### Change description ###
Addresses updates to change "addresses" to "Email" in relevant configuration files and code. This includes updating labels, field names, and documentation to ensure consistency and clarity regarding email fields.

!!!! Please be aware changing field type in ccd may need migration for existing cases and elastic search may need reindexing !!!

```
[ ] Yes
[ ] No
```


tom-saunders-cts:
this doesn't capture all the fields we currently use to store email addresses - it's also likely that if we did update those fields to use the email type as here we would then have cases with invalid emails (we know that the regex we currently use is incorrect) - since these are data about other parties (solicitors or citizens) we may not be able to reasonably correct them if they are invalid

also if the validation used in the ccd types documentation on confluence is correct (and it seems it may not be) the regex there does not come close to correctly handling email addresses

PR #3560, opened 2025-12-12
DTSPB-4952 Dummy Postcode for Invalid postCode

MarcusHMCTS:

[DTSPB-4952](https://tools.hmcts.net/jira/browse/DTSPB-4952)

### Change description ###

Bulkscan: Set dummy value for invalid/foreign postcode

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/probate/service/exceptionrecord/mapper/OCRFieldAddressMapper.java
+++ b/src/main/java/uk/gov/hmcts/probate/service/exceptionrecord/mapper/OCRFieldAddressMapper.java
@@ -3,7 +3,6 @@ package uk.gov.hmcts.probate.service.exceptionrecord.mapper;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
 import org.springframework.stereotype.Component;
-import uk.gov.hmcts.probate.config.BulkScanConfig;
 import uk.gov.hmcts.probate.exception.OCRMappingException;
 import uk.gov.hmcts.probate.model.exceptionrecord.ExceptionRecordOCRFields;
 import uk.gov.hmcts.probate.service.exceptionrecord.mapper.qualifiers.ToAttorneyOnBehalfOfAddress;
@@ -37,12 +36,6 @@ public class OCRFieldAddressMapper {
     private static final String DECEASED_ADDRESS_POSTCODE = "deceasedAddressPostCode";
     private static final String CAVEATOR_ADDRESS_POSTCODE = "caveatorAddressPostCode";
 
-
-    private final BulkScanConfig bulkScanConfig;
-
-    public OCRFieldAddressMapper(BulkScanConfig bulkScanConfig) {
-        this.bulkScanConfig = bulkScanConfig;
-    }
 
     @SuppressWarnings("squid:S1168")
     @ToPrimaryApplicantAddress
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/hmcts/probate/service/exceptionrecord/mapper/OCRFieldAddressMapper.java
```diff
@@ -35,6 +38,9 @@ public class OCRFieldAddressMapper {
     private static final String DECEASED_ADDRESS_POSTCODE = "deceasedAddressPostCode";
     private static final String CAVEATOR_ADDRESS_POSTCODE = "caveatorAddressPostCode";
 
+    @Autowired
+    BulkScanConfig bulkScanConfig;
```
tom-saunders-cts: I think it's clearer to make dependencies `private final` and inject them via the constructor - means that tests can manually constructor inject rather than relying on reflection to do so as in the testing below.

src/main/java/uk/gov/hmcts/probate/service/exceptionrecord/mapper/OCRFieldAddressMapper.java
```diff
@@ -159,9 +165,12 @@ private Address buildAddress(String postCodeField) {
...
         if (StringUtils.isBlank(address.getPostCode())) {
             return null;
-        } else {
-            address.setPostCode(postCode.toUpperCase());
+        }
+        address.setPostCode(postCode.toUpperCase());
+        try {
             validatePostCode(address.getPostCode(), postCodeField);
+        } catch (OCRMappingException e) {
+            address.setPostCode(bulkScanConfig.getPostcode());
```
tom-saunders-cts: do we want to log this having happened? (it seems like `validatePostCode` is doing so but it's logging an error with no real way to identify which case etc.)

is there anything to report to a user that this happened? it seems like the 'normal' behaviour here would be that the exception we are catching would end up being reported to the caller but that is now not happening

src/test/java/uk/gov/hmcts/probate/service/exceptionrecord/mapper/ExceptionRecordCaveatMapperAfterMappingTest.java
tom-saunders-cts, line 1: unclear to me why we need this (or the other exceptionrecord file changes)

src/test/java/uk/gov/hmcts/probate/service/exceptionrecord/mapper/OCRFieldAddressMapperTest.java
```diff
@@ -43,8 +49,16 @@ class OCRFieldAddressMapperTest {
...
+    @Mock
+    private BulkScanConfig bulkScanConfig;
+
     @BeforeEach
     public void setUpClass() throws Exception {
+        MockitoAnnotations.openMocks(this);
+        when(bulkScanConfig.getPostcode()).thenReturn(DEFAULT_POSTCODE_VALUE);
+        Field bulkScanConfigField = OCRFieldAddressMapper.class.getDeclaredField("bulkScanConfig");
+        bulkScanConfigField.setAccessible(true);
+        bulkScanConfigField.set(addressMapper, bulkScanConfig);
```
tom-saunders-cts: as mentioned above if we pass this through the constructor this becomes simpler

src/main/java/uk/gov/hmcts/probate/service/exceptionrecord/mapper/OCRFieldAddressMapper.java
MarcusHMCTS: updated, please check again

tom-saunders-cts (dismissed):
this still feels a little indirect but as best i can understand it should achieve the request made in the ticket.

tom-saunders-cts:
one question i have here - what is the behaviour when the postcode is replaced like this? is the original value presented to caseworkers somewhere?

PR #3566, opened 2025-12-15
DTSPB-5113 migrate dormant case to previous state if were in post grant states

FeliTam:
[DTSPB-5113](https://tools.hmcts.net/jira/browse/DTSPB-5113)

### Change description ###
Migrate dormant case to previous state if were in post grant states

```
[ ] Yes
[x] No
```


tom-saunders-cts (dismissed):
i think this should do what we expect here, just need to check the previous state etc.

PR #3582, opened 2025-12-23
Dtspb 5137 remove caveat stop reasons

IswaryaPepakayala:

https://tools.hmcts.net/jira/browse/DTSPB-5137

### Change description ###

```
[ ] Yes
[X] No
```


FeliTam (commented):
the changes overall look sensible to me
Could we add the filtering to the ES query, for example:
`"must_not": [
{
          "terms": {
            "data.boCaseStopReasonList.value.caseStopReason.keyword": [
              "CaveatMatch",
              "Permanent Caveat"
            ]
          }
        }
      ]
`
This would prevent those cases from being fetched in the first place.

FeliTam (approved):

PR #3583, opened 2025-12-31
DTSPB-5053 Add deceased name and dod to disposalReminder and unsubmittedApplication notifications

FeliTam:
[DTSPB-5053](https://tools.hmcts.net/jira/browse/DTSPB-5053)

### Change description ###
Add deceased name and dod to disposalReminder and unsubmittedApplication notifications

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/probate/service/notification/AutomatedNotificationPersonalisationService.java
```diff
@@ -75,46 +85,49 @@ public AutomatedNotificationPersonalisationService(
...
-        HashMap<String, String> personalisation = new HashMap<>();
+        log.info("Building DisposalReminder personalisation for caseId={}, applicationType={}",
+                caseDetails.getId(), applicationType);
+        HashMap<String, Object> personalisation = new HashMap<>();
         Map<String, Object> data = caseDetails.getData();
-        personalisation.put(PERSONALISATION_DATE_CREATED, DATE_FORMAT.format(caseDetails.getCreatedDate()));
-        personalisation.put(PERSONALISATION_CASE_ID, caseDetails.getId().toString());
+
+        addCreatedDates(personalisation, caseDetails.getCreatedDate());
+        addDeceasedDetails(personalisation, data);
```
tom-saunders-cts: i know that having this functionality split out so it's not repeated is generally a good idea but i think there is value in having the personalisation methods actually list the fields which are being populated - it makes it simpler to confirm that the values provided match those expected by Notify

maybe a compromised would be that we have a method which returns on object containing the values (so that the generation for the welsh isn't repeated) but then we set the values explicitly per message type. this also means that where we are (already) inconsistent with the naming of parameters we can adapt

src/main/resources/application.yml
```diff
@@ -391,7 +391,7 @@ notifications:
           hseReminder: "38e4f0e6-5656-4ae9-82f7-ac3da178100f"
           hseReminderForHub: "f039720f-8bdb-48f8-9549-f0068d7a5688"
           dormantWarning: "9cfa645e-4849-433e-8b72-382a27d7ddea"
-          unsubmittedApplication: "3f3965ca-1c8a-4b44-be66-75b929513e65"
+          unsubmittedApplication: "168c520d-3008-4584-8b59-865debb86fa7"
```
tom-saunders-cts, line 398: the text of the updated notification is (emphasis mine):
> Deceased’s name: ((deceased_name))
> 
> Deceased’s date of death: ((deceased_dod))
> 
> You started an online probate application on ((date_created)) but have not yet submitted it.
> 
> **Deceased’s name: Name not entered yet**
> **Deceased’s date of death: Date not entered yet**
> 
> Caseworkers cannot start working on your application until you have submitted it and made payment.

src/main/resources/application.yml
```diff
@@ -499,14 +499,14 @@ notifications:
           grantDelayed: "7045b381-a087-44fe-8bb5-e5a591cb8473"
           grantAwaitingDocumentation: "638395bf-805f-48a7-97b1-d26ea9e9334e"
           noticeOfChangeReceived: "4f1f3c47-1d84-4584-b251-2a77d64510f2"
-          disposalReminder: "bd041f52-5f87-4d90-87ab-302e20e7215d"
-          caveatDisposalReminder: "f5268c11-f0eb-436a-a8f8-df3a85519c2d"
+          disposalReminder: "f9de5515-17f0-4e62-a11f-07357add1ca6"
+          caveatDisposalReminder: "760335be-a158-4006-8241-014becb2065c"
```
tom-saunders-cts: this message is only in english (both before and after (?)

src/main/resources/application.yml
FeliTam: updated

src/main/resources/application.yml
FeliTam: Left a comment on the ticket to confirm with the ST

src/main/resources/application.yml
```diff
@@ -499,14 +499,14 @@ notifications:
...
-          caveatDisposalReminder: "f5268c11-f0eb-436a-a8f8-df3a85519c2d"
+          disposalReminder: "f9de5515-17f0-4e62-a11f-07357add1ca6"
+          caveatDisposalReminder: "501f91bb-dcce-4001-afac-6644b23b3acb"
           draftCasePaymentSuccess: "b881e8ea-76a5-412c-83d5-8e0f08791b16"
           firstStopReminder: "e06d92b6-e59f-4a78-8654-9c1a9f70411b"
           secondStopReminder: "206a11a2-1171-464b-ad03-5ca83b928968"
           hseReminder: "0146b124-a67c-4753-b38e-85bcb37d8562"
           dormantWarning: "f2d04d10-fb1e-4693-a3d7-845a64cfcc7c"
-          unsubmittedApplication: "7440ee74-4cde-4cd4-ab76-038d70674424"
+          unsubmittedApplication: "1f67e261-c7f7-41f7-9c86-46fe2a2bb118"
```
tom-saunders-cts, line 513: all the other templates here appear to use a `welsh_deceased_name` for the name of the deceased but this just use `deceased_name` which i guess means that if no name has yet been provided the text in the welsh email will end up being:

> Enw'r ymadawedig: Name not entered yet
rather than
> Enw'r ymadawedig: Enw heb ei nodi eto

I'm not sure if that's intentional or just an oversight

src/main/java/uk/gov/hmcts/probate/service/notification/AutomatedNotificationPersonalisationService.java
```diff
@@ -75,46 +84,69 @@ public AutomatedNotificationPersonalisationService(
...
+                localDateToWelshStringConverter.convert(caseDetails.getCreatedDate().toLocalDate()));
         personalisation.put(TODAY, LocalDate.now().format(DATE_FORMAT));
         personalisation.put(TODAY_WELSH, localDateToWelshStringConverter.convert(LocalDate.now()));
-        personalisation.put(PERSONALISATION_DATE_CREATED, DATE_FORMAT.format(caseDetails.getCreatedDate()));
-        personalisation.put(PERSONALISATION_CCD_REFERENCE, caseDetails.getId().toString());
-        personalisation.put(PERSONALISATION_RESPOND_DATE, DATE_FORMAT.format(LocalDate.now().plusDays(14)));
+        personalisation.put(PERSONALISATION_CCD_REFERENCE, String.valueOf(caseDetails.getId()));
+        LocalDate respondBy = LocalDate.now().plusDays(RESPOND_BY_DAYS);
+        personalisation.put(PERSONALISATION_RESPOND_DATE, respondBy.format(DATE_FORMAT));
+        personalisation.put(PERSONALISATION_RESPOND_DATE_WELSH, localDateToWelshStringConverter.convert(respondBy));
```
tom-saunders-cts, line 139: i can't actually find a template (at least a template amended in this PR) which uses this value? (this is one of the major downsides of having one do-all method to get personalisation for all templates - it's hard to reason about whether the correct values are actually sent and whether they're actually used - see comment on the template config for the other side of this)

src/main/resources/application.yml
MarcusHMCTS: thanks, updated to welsh_deceased_name

tom-saunders-cts (dismissed):
i think this is now passing the correct values into the templates, but as discussed i'm not sure that having a single do-everything method for our template injection actually helps us.

it's not a thing we should look at here, but i wonder whether removing the split handling for our bilingual templates and just always sending a template with english text above welsh (with a note in welsh that welsh text is below) might be a simpler to maintain solution

MarcusHMCTS (approved):

PR #3587, opened 2026-01-02
DTSPB-5057 hide Caveat not matched event

FeliTam:
DTSPB-5057

### Change description ###
hide Caveat not matched event

```
[ ] Yes
[x] No
```


ccdImports/configFiles/CCD_Probate_Caveat/AuthorisationCaseEvent.json
```diff
@@ -509,6 +509,7 @@
   {"CaseTypeID": "Caveat", "CaseEventID": "handleEvidence", "UserRole": "caseworker-probate-systemupdate", "CRUD": "CRU"},
   {"CaseTypeID": "Caveat", "CaseEventID": "raiseCaveatFromBulkScan", "UserRole": "caseworker-probate-systemupdate", "CRUD": "CRU"},
   {"CaseTypeID": "Caveat", "CaseEventID": "createCase", "UserRole": "caseworker-probate-systemupdate", "CRUD": "CRU"},
+  {"CaseTypeID": "Caveat", "CaseEventID": "caveatNotMatched", "UserRole": "caseworker-probate-systemupdate", "CRUD": "CRUD"},
```
tom-saunders-cts, line 517: if we are adding this do we expect it to actually be used (i.e. should this be only read as well)

tom-saunders-cts (dismissed):
i remain unconvinced that the change to the end state is actually correct in the wider context, but it is what has been requested - it won't break things but i think it'll just mean cases can go in cycles

PR #3591, opened 2026-01-05
DTSPB-5087 added stop code docs required - PA1A and PA1P

justiceia:
https://tools.hmcts.net/jira/browse/DTSPB-5087

### Change description ###
adding to stop case documents required PA1A and PA1P including changes to 8 and 12 week reminder

```
[ ] Yes
[x] No
```


tom-saunders-cts (approved):
seems reasonable

the only question i have is whether the conversion from value as used internally to externally should have details on what these are. i don't think that's explicitly asked for so it's probably fine without, but if i recall that's used to send a list of reasons that a case is stopped so it may be useful for there being a little more context for someone who doesn't live and breathe probate application form names

PR #3673, opened 2026-03-02
DTSPB-5156 Update S&F manifest

jamiealbertelli:
[DTSPB-5156](https://tools.hmcts.net/jira/browse/DTSPB-5156)

### Change description ###
Update S&F manifest to include comment field and fix subtype and casetype mismatch

- [ ] Yes
- [x] No


src/main/java/uk/gov/hmcts/probate/service/zip/ZipFileService.java
```diff
@@ -74,7 +78,9 @@ public class ZipFileService {
         ADMON_WILL_GRANT_REISSUE, AD_COLLIGENDA_BONA_GRANT_REISSUE, WELSH_DIGITAL_GRANT_REISSUE,
         WELSH_INTESTACY_GRANT_REISSUE, WELSH_ADMON_WILL_GRANT_REISSUE, WELSH_AD_COLLIGENDA_BONA_GRANT_REISSUE};
     private static final String HEADER_ROW_FILE = "templates/dataExtracts/ManifestFileHeaderRow.csv";
-    private static final String ERROR_MASSAGE = "Exception occurred while generating zip file ";
+    private static final String HEADER_ROW_FILE_WITHOUT_COMMENT =
```
tom-saunders-cts: these are now not used since the header is manually put in place as the first record

src/main/java/uk/gov/hmcts/probate/service/zip/ZipFileService.java
```diff
@@ -280,32 +301,41 @@ private void zipMultipleDocs(final ZipOutputStream zos,
...
-    }
-
-    private void generateManifestFile(ZipOutputStream zos, List<ZippedManifestData> zippedManifestDataList)
+    private void generateManifestFile(ZipOutputStream zos, List<ZippedManifestData> zippedManifestDataList,
+                                      boolean shouldIncludeComment)
             throws IOException {
         StringBuilder data = new StringBuilder();
-        addHeaderRow(data);
-        data.append(NEW_LINE);
+        try (final CSVPrinter csvWriter = new CSVPrinter(data, CSVFormat.RFC4180)) {
```
tom-saunders-cts: RFC4180 will mean we send the file with records terminated by CRLF (`\r\n`) where we have previously only been sending LF (`\n`) - are we confident this is not going to cause issues for the recipients?

tom-saunders-cts: we may be 'better' placed building a specific CSVFormat which matches the current behaviour (not quoting unless needed due to included commas, `\r` record separator

src/main/java/uk/gov/hmcts/probate/service/zip/ZipFileService.java
```diff
@@ -280,32 +301,41 @@ private void zipMultipleDocs(final ZipOutputStream zos,
...
-            data.append(zippedManifestData.getDocumentName());
-            data.append(DELIMITER);
-            data.append(zippedManifestData.getErrorDescription());
-            data.append(NEW_LINE);
+            for (ZippedManifestData zippedManifestData : zippedManifestDataList) {
+                csvWriter.print(zippedManifestData.getCaseNumber());
+                csvWriter.print(zippedManifestData.getDocumentId());
+                csvWriter.print(zippedManifestData.getDocType());
+                csvWriter.print(zippedManifestData.getSubType());
+                csvWriter.print(zippedManifestData.getCaseType());
```
tom-saunders-cts, line 346: we are maintaining the order here with the previous behaviour, altering the order in the headers based on the flag but also correcting the values stored (at line 238) - i think this means that both forms will now be incorrect?

src/main/java/uk/gov/hmcts/probate/service/zip/ZipFileService.java
```diff
@@ -78,8 +79,6 @@ public class ZipFileService {
         ADMON_WILL_GRANT_REISSUE, AD_COLLIGENDA_BONA_GRANT_REISSUE, WELSH_DIGITAL_GRANT_REISSUE,
         WELSH_INTESTACY_GRANT_REISSUE, WELSH_ADMON_WILL_GRANT_REISSUE, WELSH_AD_COLLIGENDA_BONA_GRANT_REISSUE};
     private static final String HEADER_ROW_FILE = "templates/dataExtracts/ManifestFileHeaderRow.csv";
-    private static final String HEADER_ROW_FILE_WITHOUT_COMMENT =
```
tom-saunders-cts: i think these are both unused? or have i missed something using HEADER_ROW_FILE somewhere

jamiealbertelli: Yep removed

src/test/java/uk/gov/hmcts/probate/service/zip/ZipFileServiceTest.java
```diff
@@ -107,6 +107,10 @@ void setUp() throws Exception {
         when(fileSystemResourceService.getFileFromResourceAsString("templates/dataExtracts/ManifestFileHeaderRow.csv"))
                 .thenReturn("Case reference number|Document id|Document type|"
                         + "Document sub type|Case type|Document file name|Error description");
+        when(fileSystemResourceService.getFileFromResourceAsString(
+                "templates/dataExtracts/ManifestFileHeaderRowWithoutComment.csv"))
+                .thenReturn("Case reference number|Document id|Document type|Document sub type"
+                        + "|Case type|Document file name|Error description");
         when(smeeAndFordPersonalisationService.getSmeeAndFordByteArray(anyList()))
```
tom-saunders-cts, line 114: i'm not entirely convinced these are ever called(?)

tom-saunders-cts (dismissed):
only query is whether the methods added in ZipFileService should be non-static - not going to hold up on that though

FeliTam (approved):

PR #3680, opened 2026-03-03
fix createAllXLS.sh

FeliTam:
### Change description ###
fix createAllXLS.sh to correctly accept the shutterOption param ${3}

```
[ ] Yes
[x] No
```

### DIFF
--- a/ccdImports/conversionScripts/createAllXLS.sh
+++ b/ccdImports/conversionScripts/createAllXLS.sh
@@ -4,8 +4,14 @@ set -eu
 
 conversionFolder=$(dirname "$0")
 configFolder=${conversionFolder}/../configFiles
-shutterOption=${2:-false}
-extraExclusions=${3:-",*-wa.json"}
+# Params:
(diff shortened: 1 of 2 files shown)


ccdImports/conversionScripts/createAllXLS.sh
tom-saunders-cts: if we're naming our inputs let's use the names consistently?

tom-saunders-cts: i'm also not sure this test is logically the right thing - i think we should probably be checking we have at least two arguments (since the first two are required)

tom-saunders-cts (changes requested):

ccdImports/conversionScripts/createAllXLS.sh
FeliTam: updated

MarcusHMCTS (approved):

tom-saunders-cts (approved):

PR #3693, opened 2026-03-11
DTSPB-5162 change search size to 100 and adding pagination and lots of refactoring

justiceia:
https://tools.hmcts.net/jira/browse/DTSPB-5162

### Change description ###
changed query to use 100 as size with pagination instead of the ES default of 10

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/probate/service/CaveatQueryService.java
```diff
@@ -134,7 +136,22 @@ public List<ReturnedCaveatDetails> findCaveatExpiredCases(String expiryDate) {
...
+            SearchSourceBuilder sourceBuilder = new SearchSourceBuilder()
+                    .query(query)
+                    .sort(SORT_COLUMN, SortOrder.ASC)
+                    .from(from)
+                    .size(dataExtractPaginationSize);
+            String jsonQuery = sourceBuilder.toString();
+            pageResults = runQuery(CAVEAT, jsonQuery).getCaveats();
+            allResults.addAll(pageResults);
+            from += dataExtractPaginationSize;
+        } while (!pageResults.isEmpty());
```
FeliTam: This would cause an unnecessary empty-page call.
maybe we can use `while (pageResults.size() == dataExtractPaginationSize);`

FeliTam (changes requested):

src/main/java/uk/gov/hmcts/probate/service/CaveatQueryService.java
FeliTam: It seems the pagination should process the fetched cases before querying ES again, similar to the approach used in the automated notification job. Otherwise, there is a risk of an OOM exception if all 6–7k cases are fetched before processing the expiry events.

src/main/java/uk/gov/hmcts/probate/service/CaveatQueryService.java
```diff
@@ -91,50 +99,50 @@ public CaveatData findCaveatById(CaseType caseType, String caveatId) {
...
 
+    public List<ReturnedCaveatDetails> fetchExpiredCaveatsPage(String expiryDate, Long[] searchAfterValues) {
+        BoolQueryBuilder query = buildExpiryQuery(expiryDate);
+        SearchSourceBuilder sourceBuilder = new SearchSourceBuilder()
+                .query(query)
+                .sort(SORT_COLUMN, SortOrder.ASC)
+                .size(dataExtractPaginationSize);
+        if (searchAfterValues != null) {
+            sourceBuilder.searchAfter(searchAfterValues);
+        }
```
tom-saunders-cts, line 114: should this not be filtering the response to only include the case reference? (rather than the full case data?)

src/main/java/uk/gov/hmcts/probate/service/caveat/CaveatExpiryServiceImpl.java
```diff
@@ -28,52 +26,61 @@
...
-                    failedCases);
-        }
+        Long[] searchAfter = null;
+        List<ReturnedCaveatDetails> pageResults;
+        do {
+            pageResults = caveatQueryService.fetchExpiredCaveatsPage(expiryDate, searchAfter);
+            log.info("Processing {} caveats in current page", pageResults.size());
+            SecurityDTO securityDTO = securityUtils.getSecurityDTO();
+            for (ReturnedCaveatDetails caveat : pageResults) {
+                expireCaveat(caveat, securityDTO, failedCases);
```
tom-saunders-cts: why have we changed this from gathering all the candidate cases first and then expiring them to expiring them whilst searching ?

tom-saunders-cts: or i guess why do we have the looping in the expiry serivce rather than in the query service

src/main/java/uk/gov/hmcts/probate/service/caveat/CaveatExpiryServiceImpl.java
```diff
@@ -28,52 +26,61 @@
...
+        uk.gov.hmcts.reform.probate.model.cases.caveat.CaveatData caseData =
+                uk.gov.hmcts.reform.probate.model.cases.caveat.CaveatData.builder()
+                        .autoClosedExpiry(Boolean.TRUE)
+                        .build();
         try {
-            ccdClientApi
-                    .updateCaseAsCaseworker(CcdCaseType.CAVEAT, caseId, lastModified, caseData, eventIdToStart,
-                            securityDto, EVENT_DESCRIPTOR_CAVEAT_EXPIRED, EVENT_DESCRIPTOR_CAVEAT_EXPIRED);
-            log.info("Caveat autoExpired: {}", caseId);
+            ccdClientApi.updateCaseAsCaseworker(
```
tom-saunders-cts: this doesn't meet the expectations for we should be handling concurrent ccd events - we should be:
- find candidates for auto closure from elasticsearch
- for each candidate:
  - start the relevant event
  - check the case data & state returned as part of the start event response
  - if the current case state matches our expecations:
    - mutate the case data returned in the start event response
    - submit the mutated case data

doing what we are doing here runs the (admittedly small) risk that something happens to update the case then our autoclosure could overwrite some of the changes made to the case data.

in practic (specifically here)  i think this isn't actually a problem because all we are actually providing in the response data is the autoClosedExpiry boolean, but i think this only works safely because the ccd event api doesn't allow a caller to delete fields

src/main/java/uk/gov/hmcts/probate/service/caveat/CaveatExpiryServiceImpl.java
justiceia: there was an issue where it could get an out of memory if we got them all first

src/main/java/uk/gov/hmcts/probate/service/CaveatQueryService.java
justiceia: i just used the id because it was there but not being used i thought someone may have tried to do it initially. wouldnt that take a while to sort if we sorted the entire case data?

tom-saunders-cts (dismissed):
only question is whether we should have a test which confirms that we successfully loop when we have more than pageSize results (unless I've missed something here and we do?)

justiceia:
> only question is whether we should have a test which confirms that we successfully loop when we have more than pageSize results (unless I've missed something here and we do?)

changed it a bit

tom-saunders-cts (dismissed):

build.gradle
```diff
@@ -262,9 +265,11 @@ dependencies {
   annotationProcessor group:'org.projectlombok', name:'lombok', version: versions.lombok
   implementation group: 'org.projectlombok', name: 'lombok-mapstruct-binding', version: '0.2.0'
 
-  implementation group: 'org.elasticsearch.client', name: 'elasticsearch-rest-high-level-client', version: versions.elasticSearchVersion
   implementation group: 'org.elasticsearch.client', name: 'elasticsearch-rest-client', version: versions.elasticSearchVersion
   implementation group: 'org.elasticsearch', name: 'elasticsearch', version: versions.elasticSearchVersion
+
+  implementation group: 'joda-time', name: 'joda-time', version: '2.12.6'
```
tom-saunders-cts: why are we pulling in joda-time? we're long past java 8?

justiceia: it was needed for some payment related tests that was removed in elasticsearch-rest-high-level-client

src/main/java/uk/gov/hmcts/probate/service/CaveatQueryService.java
justiceia: changed

MarcusHMCTS (dismissed):

src/main/java/uk/gov/hmcts/probate/service/CaveatQueryService.java
```diff
@@ -109,7 +109,8 @@ public List<ReturnedCaveatDetails> fetchExpiredCaveatsPage(String expiryDate, Lo
         if (searchAfterValues != null) {
             sourceBuilder.searchAfter(searchAfterValues);
         }
-        String jsonQuery = sourceBuilder.toString();
+        String jsonQuery = sourceBuilder.toString()
+                .replaceFirst("\\{", "{\"_source\":[\"reference\"],");
```
tom-saunders-cts: this feels like the wrong way to handle this

tom-saunders-cts (commented):
only question is whether we should document _why_ this weird:
`queryBuilder.toString()` -> `JSON` -> modify -> `toString()` dance

(since otherwise someone is likely to try and just use the built in functionality as was originally suggested)

tom-saunders-cts (dismissed):
only question is whether we should document _why_ this weird:
`queryBuilder.toString()` -> `JSON` -> modify -> `toString()` dance

(since otherwise someone is likely to try and just use the built in functionality as was originally suggested)

tom-saunders-cts (approved):

FeliTam (approved):

PR #3704, opened 2026-03-16
Dtspb 4993 allow cw attach docs

IswaryaPepakayala:

https://tools.hmcts.net/jira/browse/DTSPB-4993

### Change description ###

```
[ ] Yes
[X] No
```


ccdImports/configFiles/CCD_Probate_Backoffice/AuthorisationCaseField/AuthorisationCaseField.json
```diff
@@ -4045,6 +4069,8 @@
   {"CaseTypeID": "GrantOfRepresentation", "CaseFieldID": "informationNeededByPostSwitch", "UserRole": "caseworker-probate-rparobot", "CRUD": "CRUD"},
   {"CaseTypeID": "GrantOfRepresentation", "CaseFieldID": "emailPreview", "UserRole": "caseworker-probate-rparobot", "CRUD": "CRUD"},
   {"CaseTypeID": "GrantOfRepresentation", "CaseFieldID": "boDocumentsUploaded", "UserRole": "caseworker-probate-rparobot", "CRUD": "CRUD"},
+  {"CaseTypeID": "GrantOfRepresentation", "CaseFieldID": "cwDocumentUpload", "UserRole": "caseworker-probate-rparobot", "CRUD": "CRUD"},
+  {"CaseTypeID": "GrantOfRepresentation", "CaseFieldID": "cwDocumentUploadedList", "UserRole": "caseworker-probate-rparobot", "CRUD": "CRUD"},
```
tom-saunders-cts: do we expect the robotic automation to be using this?

ccdImports/configFiles/CCD_Probate_Backoffice/CaseTypeTab.json
```diff
@@ -7,8 +7,9 @@
...
   {"LiveFrom": "01/01/2017", "CaseTypeID": "GrantOfRepresentation", "Channel": "CaseWorker", "TabID": "documentsTab", "TabLabel": "Documents", "TabDisplayOrder": 4, "CaseFieldID": "probateDocumentsGenerated", "TabFieldDisplayOrder": 1},
   {"LiveFrom": "02/01/2017", "CaseTypeID": "GrantOfRepresentation", "Channel": "CaseWorker", "TabID": "documentsTab", "TabLabel": "Documents", "TabDisplayOrder": 4, "CaseFieldID": "boDocumentsUploaded", "TabFieldDisplayOrder": 2},
-  {"LiveFrom": "01/01/2017", "CaseTypeID": "GrantOfRepresentation", "Channel": "CaseWorker", "TabID": "documentsTab", "TabLabel": "Documents", "TabDisplayOrder": 4, "CaseFieldID": "scannedDocuments", "TabFieldDisplayOrder": 3},
-  {"LiveFrom": "01/01/2017", "CaseTypeID": "GrantOfRepresentation", "Channel": "CaseWorker", "TabID": "documentsTab", "TabLabel": "Documents", "TabDisplayOrder": 4, "CaseFieldID": "citizenResponses", "TabFieldDisplayOrder": 4},
+  {"LiveFrom": "02/01/2017", "CaseTypeID": "GrantOfRepresentation", "Channel": "CaseWorker", "TabID": "documentsTab", "TabLabel": "Documents", "TabDisplayOrder": 4, "CaseFieldID": "cwDocumentUploadedList", "TabFieldDisplayOrder": 3},
```
tom-saunders-cts, line 10: is this field being present here (top of the documents tab) explicitly requested in the ticket - it seems reasonable that these should be available somewhere but i wonder if maybe on the notifications tab might make more sense?

src/main/java/uk/gov/hmcts/probate/model/ccd/raw/response/ResponseCaseData.java
```diff
@@ -421,4 +421,7 @@ public class ResponseCaseData extends ResponseCaseDataParent {
     private final String executorsNamed;
     private final LocalDate firstStopReminderSentDate;
     private String issueEarlySwitch;
+    private final String uploadFileCheck;
+    private final UploadDocument cwDocumentUpload;
```
tom-saunders-cts: do we want/need these in the response (does it mean that they will be pre-populated the next time someone sends a RFI?)

src/main/java/uk/gov/hmcts/probate/security/SecurityUtils.java
```diff
@@ -173,6 +173,7 @@ public void setSecurityContextUserAsScheduler() {
     }
 
     public String getCaseworkerToken() {
+        log.info("Getting caseworker token for user: {} {}", caseworkerUserName, caseworkerPassword);
```
tom-saunders-cts: we cannot be logging these values

src/main/java/uk/gov/hmcts/probate/service/NotificationService.java
```diff
@@ -208,22 +216,71 @@ public Document emailPreview(CaseDetails caseDetails) throws NotificationClientE
...
+        } else if (NO.equalsIgnoreCase(caseData.getUploadFileCheck())) {
+            personalisation.put("link_to_file", " ");
+            personalisation.put("expiry_date", " ");
+        }
+    }
+
+    private void addExpiryDatePersonalisation(Map<String, Object> personalisation) {
+
+        DateTimeFormatter expiryFormatter = DateTimeFormatter.ofPattern("dd/MM/yyyy");
+        LocalDate expiryDate = LocalDate.now().plusWeeks(26);
```
tom-saunders-cts: we should probably use a single constant for both this instance of 26 and the value used in `cwPrepareUpload` (i don't know that it needs to be configurable but it saves us from missing updating one of these if we need to in the future)

src/main/java/uk/gov/hmcts/probate/service/NotificationService.java
```diff
@@ -208,22 +216,71 @@ public Document emailPreview(CaseDetails caseDetails) throws NotificationClientE
...
+        personalisation.put("expiry_date", expiryDate.format(expiryFormatter));
+    }
+
+    private void addCwDocumentToPersonalisation(
```
tom-saunders-cts, line 265: both this and `cwPrepareUpload` have error conditions which we catch (and log, so they're not lost) but then continue processing which probably leads to unexpected behaviour for caseworkers if they are expecting the link to actually be included.

src/main/java/uk/gov/hmcts/probate/service/NotificationService.java
```diff
@@ -208,22 +216,71 @@ public Document emailPreview(CaseDetails caseDetails) throws NotificationClientE
...
-        String templateId = templateService.getTemplateId(CASE_STOPPED_REQUEST_INFORMATION,
-            caseData.getApplicationType(), caseData.getRegistryLocation(), caseData.getLanguagePreference(),
-                null, caseData.getChannelChoice(), caseData.getInformationNeededByPost());
         Map<String, Object> personalisation =
                 grantOfRepresentationPersonalisationService.getPersonalisation(caseDetails,
                         registry);
+        personalisation.put("upload_check", caseData.getUploadFileCheck());
 
         updatePersonalisationForSolicitor(caseData, personalisation);
+        addPersonalisationForUploadDocument(caseData, personalisation);
```
tom-saunders-cts: does this actually end up with a link in the preview response? presumably if so it's never going to be accessible to anyone since the link will be regenerated when sent

src/main/java/uk/gov/hmcts/probate/transformer/CallbackResponseTransformer.java
```diff
@@ -1444,7 +1460,10 @@ private CallbackResponse transformResponse(ResponseCaseData responseCaseData) {
             .executorsNamed(caseData.getExecutorsNamed())
             .ttl(caseData.getTtl())
             .firstStopReminderSentDate(caseData.getFirstStopReminderSentDate())
-            .evidenceHandledDate(caseData.getEvidenceHandledDate());
+            .evidenceHandledDate(caseData.getEvidenceHandledDate())
+            .uploadFileCheck(caseData.getUploadFileCheck())
+            .cwDocumentUpload(caseData.getCwDocumentUpload())
```
tom-saunders-cts: do we actually expect these values to be persisted into the case data?

tom-saunders-cts (changes requested):

src/main/java/uk/gov/hmcts/probate/transformer/CallbackResponseTransformer.java
IswaryaPepakayala: removed.

src/main/java/uk/gov/hmcts/probate/security/SecurityUtils.java
IswaryaPepakayala: removed.Sorry I added that for testing.Forgot to remove

src/main/java/uk/gov/hmcts/probate/service/NotificationService.java
IswaryaPepakayala: throwing new exception for exception cases

src/main/java/uk/gov/hmcts/probate/service/NotificationService.java
IswaryaPepakayala: added in application.yaml file

ccdImports/configFiles/CCD_Probate_Backoffice/AuthorisationCaseField/AuthorisationCaseField.json
IswaryaPepakayala: removed

src/main/java/uk/gov/hmcts/probate/model/ccd/raw/response/ResponseCaseData.java
IswaryaPepakayala: removed

src/main/java/uk/gov/hmcts/probate/service/NotificationService.java
```diff
@@ -150,6 +154,8 @@ public class NotificationService {
     private Long grantAwaitingDocumentationNotificationPeriodDays;
     @Value("${notifications.grantDelayedNotificationReleaseDate}")
     private String grantDelayedNotificationReleaseDate;
+    @Value("${upload_doc.expiry_weeks}")
+    private String expiryWeeks;
```
tom-saunders-cts, line 158: i'm always a little wary of adding a field directly configured to a service class like this - it makes it annoying to configure for testing (need to reflect over the class to make it accessible...)

another way we could do this would be to pass in a `...Config` object (or similar) which provides methods to provide the values used here which could then be mocked etc for testing, but this class already has so many dependencies

src/main/java/uk/gov/hmcts/probate/service/NotificationService.java
```diff
@@ -175,7 +181,7 @@ public Document sendEmail(State state, CaseDetails caseDetails, Optional<CaseOri
         Map<String, Object> personalisation =
             grantOfRepresentationPersonalisationService.getPersonalisation(caseDetails,
                 registry);
-        personalisation.put("upload_check", caseData.getUploadFileCheck());
+        personalisation.put(PERSONALISATION_UPLOAD_CHECK, caseData.getUploadFileCheck());
```
tom-saunders-cts: should this be in `addPersonalisationForUploadDocument`?

src/main/java/uk/gov/hmcts/probate/service/NotificationService.java
```diff
@@ -235,49 +241,59 @@ public Document emailPreview(CaseDetails caseDetails) throws NotificationClientE
...
-            Map<String, Object> personalisation) {
+            Map<String, Object> personalisation, String caseReference) {
         try {
             byte[] fileContents = documentManagementService.getDocumentByBinaryUrl(cwDocumentUpload.getDocumentLink()
                     .getDocumentBinaryUrl());
-            cwPrepareUpload(fileContents, personalisation);
+            cwPrepareUpload(fileContents, personalisation, caseReference);
         } catch (IOException e) {
             log.error("Error reading CW document file  : {}", e.getMessage());
+            final String message = MessageFormat.format("Unable reading CW document file for case : {} ",
```
tom-saunders-cts: Is this message actionable for the caseworker? (i suspect there's nothing they can do to resolve the issue themselves, but is it clear that an incident likely needs to be raised?)

src/main/java/uk/gov/hmcts/probate/service/NotificationService.java
```diff
@@ -235,49 +241,59 @@ public Document emailPreview(CaseDetails caseDetails) throws NotificationClientE
...
                             fileContents,
                             false,
-                            new RetentionPeriodDuration(26, ChronoUnit.WEEKS)
+                            new RetentionPeriodDuration(Integer.parseInt(expiryWeeks), ChronoUnit.WEEKS)
                     ));
 
         } catch (NotificationClientException e) {
             log.error("Error Preparing to send email : {} ", e.getMessage());
+            final String message = MessageFormat.format(
+                    "Unable prepare to send email for case : {} ",
```
tom-saunders-cts, line 300: looking at the implementation it seems that (at least currently) the only reason this would throw a `NotificationClientException` is that the file payload we are sending is too large - are we checking this ourselves (and passing an actionable error back to the caseworker)? otherwise i'm not sure that the message here actually helps them to self-correct any issues

src/main/java/uk/gov/hmcts/probate/service/NotificationService.java
IswaryaPepakayala: Already it has some many dependencies so don't want to add one more

src/main/java/uk/gov/hmcts/probate/service/NotificationService.java
IswaryaPepakayala: modified

src/main/java/uk/gov/hmcts/probate/service/NotificationService.java
IswaryaPepakayala: Modified error message

src/main/java/uk/gov/hmcts/probate/service/NotificationService.java
IswaryaPepakayala: Added file size check

FeliTam (dismissed):

src/main/java/uk/gov/hmcts/probate/service/NotificationService.java
```diff
@@ -272,13 +272,21 @@ private void addCwDocumentToPersonalisation(
             cwPrepareUpload(fileContents, personalisation, caseReference);
         } catch (IOException e) {
             log.error("Error reading CW document file  : {}", e.getMessage());
```
tom-saunders-cts: maybe useful to log the actual exception here so we get a stack trace in the log as well as just the message

src/integrationTest/java/uk/gov/hmcts/probate/service/NotificationServiceIT.java
```diff
@@ -1568,6 +1595,52 @@ void shouldSendEmailForRequestInformationPACtsc()
...
+                personalCaseDataCtscRequestInformation.getData().getBoStopDetailsDeclarationParagraph());
+        personalisation.put(PERSONALISATION_CASE_STOP_DETAILS,
+                personalCaseDataCtscRequestInformation.getData().getBoStopDetails());
+        personalisation.put(PERSONALISATION_CAVEAT_CASE_ID,
+                personalCaseDataCtscRequestInformation.getData().getBoCaseStopCaveatId());
+        personalisation.put(PERSONALISATION_DECEASED_DOD,
+                personalCaseDataCtscRequestInformation.getData().getDeceasedDateOfDeathFormatted());
+        personalisation.put(PERSONALISATION_CCD_REFERENCE, personalCaseDataCtscRequestInformation.getId().toString());
+        personalisation.put(PERSONALISATION_WELSH_DECEASED_DATE_OF_DEATH, localDateToWelshStringConverter
+                .convert(personalCaseDataCtscRequestInformation.getData().getDeceasedDateOfDeath()));
```
tom-saunders-cts: this doesn't appear to be used anywhere (either as an input or to compare an output against)

src/test/java/uk/gov/hmcts/probate/transformer/CallbackResponseTransformerTest.java
```diff
@@ -2686,7 +2686,12 @@ void shouldResetCitizenHubFieldsWhenHubResponseRequired() {
                 .informationNeededByPost(NO)
                 .citizenResponseCheckbox(YES)
                 .expectedResponseDate("some date")
-                .documentUploadIssue(YES);
+                .documentUploadIssue(YES)
+                .uploadFileCheck(YES)
+                .cwDocumentUpload(UploadDocument.builder()
+                        .documentLink(documentLinkMock)
+                        .documentType(SENT_EMAIL)
+                        .build());
```
tom-saunders-cts, line 2696: what does adding this `cwDocumentUpload` achieve here? we're never testing that the response has any of these values (or that it contains a collection of cw uploaded documents or similar)

FeliTam (dismissed):

FeliTam (dismissed):

FeliTam (dismissed):

FeliTam (approved):

MarcusHMCTS (approved):

PR #3709, opened 2026-03-18
DTSPB-5217 added validation for checking dob and dod

justiceia:
https://tools.hmcts.net/jira/browse/DTSPB-5217

### Change description ###
added validation for dod and dob so when exception comes in it'll throw exception if fails validation

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/probate/service/exceptionrecord/ExceptionRecordService.java
```diff
@@ -142,6 +139,11 @@ public SuccessfulTransformationResponse createGrantOfRepresentationCaseFromExcep
 
             ExceptionRecordCaseDataValidator.validateIhtValues(grantOfRepresentationData);
 
+            ExceptionRecordCaseDataValidator.validateDateOfDeath(grantOfRepresentationData);
```
justiceia: moved because i added new line which made checkstylemain fail because modifiedFields was too far away from declaration and first usage

src/main/java/uk/gov/hmcts/probate/service/exceptionrecord/utils/ExceptionRecordCaseDataValidator.java
```diff
@@ -91,6 +95,21 @@ public static void validateIhtValues(GrantOfRepresentationData caseData) {
         }
     }
 
+    public static void validateDateOfDeath(GrantOfRepresentationData caseData) {
+        List<String> errorMessages = new ArrayList<>();
+        if (caseData.getDeceasedDateOfBirth() != null && caseData.getDeceasedDateOfDeath() != null) {
```
tom-saunders-cts: so we only check if the date of death is in the future if we also have a date of birth?

src/main/java/uk/gov/hmcts/probate/service/exceptionrecord/utils/ExceptionRecordCaseDataValidator.java
```diff
@@ -91,6 +95,21 @@ public static void validateIhtValues(GrantOfRepresentationData caseData) {
...
+            if (!caseData.getDeceasedDateOfDeath().isAfter(caseData.getDeceasedDateOfBirth())) {
+                errorMessages.add(DOD_BEFORE_DOB);
+            }
+            if (caseData.getDeceasedDateOfDeath().isAfter(LocalDate.now())) {
```
tom-saunders-cts: generally prefer to use the `LocalDate.now(Clock)` implementation which allows for control of time in testing (we're sort of solving this the other way by making the test use a dynamic date of death)

src/test/java/uk/gov/hmcts/probate/service/exceptionrecord/utils/ExceptionRecordCaseDataValidatorTest.java
```diff
@@ -29,7 +31,8 @@ class ExceptionRecordCaseDataValidatorTest {
             "Net qualifying value can't be greater than the gross amount";
     private static final String IHT_NETQUALIFYING_VALUE_GREATER_THAN_ESTATE_NET_VALUE =
             "Net qualifying value can't be greater than the net amount";
-
+    private static final String DOD_BEFORE_DOB = "Date of death must be after date of birth";
+    private static final String DOD_IN_FUTURE = "Date of death cannot be in the future";
```
tom-saunders-cts: i wonder fi we should make the constants in the implementation package-scoped and then just refer to those values? it seems a little roundabout but i guess it lets us check that the error is the one we expect

src/test/java/uk/gov/hmcts/probate/service/exceptionrecord/utils/ExceptionRecordCaseDataValidatorTest.java
```diff
@@ -162,4 +165,61 @@ void shouldThrowExceptionForIhtNetQualifyingValueGraterThanNet() {
...
+    void shouldThrowExceptionWhenDeathBeforeBirth() {
+        GrantOfRepresentationData casedata = GrantOfRepresentationData.builder()
+                .deceasedDateOfBirth(LocalDate.of(2025, 1, 1))
+                .deceasedDateOfDeath(LocalDate.of(1990, 1, 1))
+                .build();
+
+        OCRMappingException exception = assertThrows(OCRMappingException.class,
+                () -> ExceptionRecordCaseDataValidator.validateDateOfDeath(casedata));
+
+        assertEquals(DOD_BEFORE_DOB, exception.getWarnings().get(0));
```
tom-saunders-cts: can we rely on this always being the 0th entry in the warning list? should we be doing something akin to
`assertThat(exception.getWarnings(), hasItem(DOD_BEFORE_DOB))` ?

src/test/java/uk/gov/hmcts/probate/service/exceptionrecord/utils/ExceptionRecordCaseDataValidatorTest.java
```diff
@@ -162,4 +165,61 @@ void shouldThrowExceptionForIhtNetQualifyingValueGraterThanNet() {
...
+    void shouldThrowExceptionWhenDeathIsSameAsBirth() {
+        LocalDate date = LocalDate.of(2024, 11, 24);
+        GrantOfRepresentationData casedata = GrantOfRepresentationData.builder()
+                .deceasedDateOfBirth(date)
+                .deceasedDateOfDeath(date)
+                .build();
+        OCRMappingException exception = assertThrows(OCRMappingException.class,
+                () -> ExceptionRecordCaseDataValidator.validateDateOfDeath(casedata));
+        assertEquals(DOD_BEFORE_DOB, exception.getWarnings().get(0));
+    }
```
tom-saunders-cts: is this test valid? it's tragic but it's not unheard of for a child to be born and then die later the same day

src/test/java/uk/gov/hmcts/probate/service/exceptionrecord/utils/ExceptionRecordCaseDataValidatorTest.java
```diff
@@ -162,4 +165,61 @@ void shouldThrowExceptionForIhtNetQualifyingValueGraterThanNet() {
...
+    @Test
+    void shouldThrowExceptionWhenDeathInFuture() {
+        GrantOfRepresentationData casedata = GrantOfRepresentationData.builder()
+                .deceasedDateOfBirth(LocalDate.now())
+                .deceasedDateOfDeath(LocalDate.now().plusDays(100))
+                .build();
+        OCRMappingException exception = assertThrows(OCRMappingException.class,
+                () -> ExceptionRecordCaseDataValidator.validateDateOfDeath(casedata));
+        assertEquals(DOD_IN_FUTURE, exception.getWarnings().get(0));
+    }
```
tom-saunders-cts, line 208: i mentioned this in the actual implementation but we should confirm the behaviour when:
```

GrantOfRepresentationData casedata = GrantOfRepresentationData.builder()
        .deceasedDateOfDeath(LocalDate.now().plusDays(100))
        .build();
```

src/test/java/uk/gov/hmcts/probate/service/exceptionrecord/utils/ExceptionRecordCaseDataValidatorTest.java
```diff
@@ -162,4 +165,61 @@ void shouldThrowExceptionForIhtNetQualifyingValueGraterThanNet() {
...
+    @Test
+    void shouldThrowMultipleErrorsForInvalidDates() {
+        GrantOfRepresentationData casedata = GrantOfRepresentationData.builder()
+                .deceasedDateOfBirth(LocalDate.now())
+                .deceasedDateOfDeath(LocalDate.now().plusDays(1))
+                .build();
+        OCRMappingException exception = assertThrows(OCRMappingException.class,
+                () -> ExceptionRecordCaseDataValidator.validateDateOfDeath(casedata));
+        assertEquals(DOD_IN_FUTURE, exception.getWarnings().get(0));
+    }
```
tom-saunders-cts, line 219: this doesn't appear to be checking for multiple errors?

tom-saunders-cts (commented):
comments inline

src/main/java/uk/gov/hmcts/probate/service/exceptionrecord/utils/ExceptionRecordCaseDataValidator.java
```diff
@@ -91,6 +95,21 @@ public static void validateIhtValues(GrantOfRepresentationData caseData) {
         }
     }
 
+    public static void validateDateOfDeath(GrantOfRepresentationData caseData) {
```
tom-saunders-cts: i'm alway wary of making methods static and returning void - it's not wrong to be clear (and is likely consistent with the existing implementation) but it can make it much more annoying to control under test conditions (vs calling on an instance of a class which we can then mock the methods of)

src/test/java/uk/gov/hmcts/probate/service/exceptionrecord/utils/ExceptionRecordCaseDataValidatorTest.java
justiceia: since we have no time value and just date, i'll assume if they're on the same day they're valid. it should not throw exception

src/main/java/uk/gov/hmcts/probate/service/exceptionrecord/utils/ExceptionRecordCaseDataValidator.java
```diff
@@ -41,10 +42,11 @@ public class ExceptionRecordCaseDataValidator {
             "Net qualifying value can't be greater than the gross amount";
     private static final String IHT_VALDIATION_ERROR = "IHT Values validation error";
     private static final String DOD_DOB_ERROR = "Date of death or date of birth error";
-    private static final String DOD_BEFORE_DOB = "Date of death must be after date of birth";
-    private static final String DOD_IN_FUTURE = "Date of death cannot be in the future";
+    static final String DOD_BEFORE_DOB = "Date of death must be after date of birth";
+    static final String DOD_IN_FUTURE = "Date of death cannot be in the future";
     private static final String INVALID_SCANNED_DOCUMENT_TYPE_ERROR = "Invalid scanned Document Type Error "
             + "for case type '%s': [%s]";
+    private static final Clock clock = Clock.systemUTC();
```
tom-saunders-cts: having this be `private static final ... Clock.systemUTC()` and passing it to `LocalDate.now(clock)` is practically the same as just using `LocalDate.now()` since we don't gain any control over the actual clock used for testing

this and the fact that the method we are using is static makes this annoying to deal with

(strictly it's different since by default it'd use `Clock.systemDefaultZone()` but it's not different in a meaningful way)

src/main/java/uk/gov/hmcts/probate/service/exceptionrecord/utils/ExceptionRecordCaseDataValidator.java
```diff
@@ -39,8 +41,12 @@ public class ExceptionRecordCaseDataValidator {
     private static final String IHT_NETQUALIFYING_VALUE_GREATER_THAN_ESTATE_GROSS_VAlUE =
             "Net qualifying value can't be greater than the gross amount";
     private static final String IHT_VALDIATION_ERROR = "IHT Values validation error";
+    private static final String DOD_DOB_ERROR = "Date of death or date of birth error";
+    static final String DOD_BEFORE_DOB = "Date of death must be after date of birth";
+    static final String DOD_IN_FUTURE = "Date of death cannot be in the future";
     private static final String INVALID_SCANNED_DOCUMENT_TYPE_ERROR = "Invalid scanned Document Type Error "
             + "for case type '%s': [%s]";
+    private final Clock clock = Clock.systemDefaultZone();
```
tom-saunders-cts: inject this through the constructor  rather than using in-class initialisation - means we can control time within the test coverage

src/main/java/uk/gov/hmcts/probate/service/exceptionrecord/utils/ExceptionRecordCaseDataValidator.java
```diff
@@ -54,11 +60,11 @@ DOC_TYPE_PPS_LEGAL_STATEMENT, singletonList(CaseType.GRANT_OF_REPRESENTATION),
                     DOC_TYPE_FORM, List.of(CaseType.GRANT_OF_REPRESENTATION, CaseType.CAVEAT)
             );
 
+    ExceptionRecordCaseDataValidator() {
```
tom-saunders-cts: this probably needs to be public and should take the Clock for injection

tom-saunders-cts (approved):

PR #3749, opened 2026-04-01
DTSPB-5045 SET HSE YES for Case Closed state

MarcusHMCTS:

[DTSPB-5045](https://tools.hmcts.net/jira/browse/DTSPB-5045)

### Change description ###

DTSPB-5045 SET HSE YES for Case Closed state

```
[ ] Yes
[x] No
```


ccdImports/configFiles/CCD_Probate_Backoffice/CaseEvent.json
```diff
@@ -182,16 +182,16 @@
...
-  {"LiveFrom": "01/01/2017", "CaseTypeID": "GrantOfRepresentation", "ID": "boWithdrawApplicationForCaseMatchingExamining", "Name": "Withdraw application", "Description": "Withdraw application", "DisplayOrder": 193, "PreConditionState(s)": "BOCaseMatchingExamining", "PostConditionState": "BOCaseClosed", "CallBackURLAboutToSubmitEvent": "http://${CCD_DEF_CASE_SERVICE_BASE_URL}/tasklist/update", "SecurityClassification": "public", "ShowSummary": "N", "ShowEventNotes": "Y"},
+  {"LiveFrom": "01/01/2017", "CaseTypeID": "GrantOfRepresentation", "ID": "boWithdrawApplicationForCasePrinted", "Name": "Withdraw application", "Description": "Withdraw application", "DisplayOrder": 184, "PreConditionState(s)": "CasePrinted", "PostConditionState": "BOCaseClosed", "CallBackURLAboutToSubmitEvent": "http://${CCD_DEF_CASE_SERVICE_BASE_URL}/tasklist//updateCaseClosed", "SecurityClassification": "public", "ShowSummary": "N", "ShowEventNotes": "Y"},
```
tom-saunders-cts: shouldn't matter but do we want the url to be `/tasklist/updateCaseClosed` with a single forward slash rather than two? (ditto below)

src/main/java/uk/gov/hmcts/probate/transformer/CaseDataTransformer.java
```diff
@@ -107,6 +108,18 @@ public void transformCaseDataForEvidenceHandled(CallbackRequest callbackRequest)
         }
     }
 
+    public void transformCaseDataForCaseCloseEvidenceHandledYes(CallbackRequest callbackRequest) {
+        if (BO_CASE_CLOSED_NAME.equals(callbackRequest.getCaseDetails().getState())) {
```
tom-saunders-cts, line 112: is the `getState()` returned here the state at the start of the event or the state after the event?

src/main/java/uk/gov/hmcts/probate/controller/DocumentController.java
```diff
@@ -369,7 +369,7 @@ public ResponseEntity<CallbackResponse> evidenceAdded(@RequestBody CallbackReque
             log.warn("evidenceAdded fails to generate or upload notification pdf for case: {}",
                     callbackRequest.getCaseDetails().getId(), e);
         }
-
+        caseDataTransformer.transformCaseDataForCaseCloseEvidenceHandledNo(callbackRequest);
```
tom-saunders-cts, line 372: we set `evidenceHandled` ->  `No` if and only if the state is `BOCaseClosed` ? 

do we not want this for other states when new evidence is added?

src/main/java/uk/gov/hmcts/probate/controller/NotificationController.java
```diff
@@ -267,6 +267,7 @@ public ResponseEntity<CallbackResponse> startDelayedNotificationPeriod(
                         callbackRequest.getCaseDetails().getId(), e);
             }
         }
+        caseDataTransformer.transformCaseDataForCaseCloseEvidenceHandledNo(callbackRequest);
```
tom-saunders-cts, line 270: why are we only setting this if we're in this state? if we want this updating don't we want it in all states?

(it seems like the evidenceHandled value is a source of great confusion which we don't really have a clear story on when it should and shouldn't be set)

src/main/java/uk/gov/hmcts/probate/controller/DocumentController.java
IswaryaPepakayala: Does it apply for boRobotUploadDocuments /evidenceAddedRPARobot
And attachScannedDocs for /notify/start-grant-delayed-notify-period

src/main/java/uk/gov/hmcts/probate/transformer/CaseDataTransformer.java
MarcusHMCTS: tested can get the state at here

ccdImports/configFiles/CCD_Probate_Backoffice/CaseEvent.json
MarcusHMCTS: fixed

FeliTam (dismissed):
Does what is mentioned in the ticket
Some endpoints are intentionally excluded from this change to avoid impacting WA configuration, which may need to be reviewed in the future

IswaryaPepakayala (approved):

FeliTam (approved):

PR #3841, opened 2026-05-13
Added flag for publish

scottiee-jc:

https://tools.hmcts.net/jira/browse/DTSPB-5631

### Change description ###

```
[ ] Yes
[ ] No
```


ccdImports/conversionScripts/createAllXLS.sh
```diff
@@ -15,6 +15,7 @@ caseServiceUrl="${1-}"
 aacUrl="${2-}"
 shutterOption="${3:-false}"
 extraExclusions="${4:-}"
+publishEnabled="${5:-false}"
```
raghera: Just wondered why you add this?

I don't think we'd have a scenario where we'd want WA to be turned on but not to publish Case events or fields.  We'd only publish events for WA as far as I know.

ccdImports/configFiles/CCD_Probate_Backoffice/CaseEvent.json
```diff
@@ -114,7 +114,7 @@
...
-  {"LiveFrom": "01/01/2017", "CaseTypeID": "GrantOfRepresentation", "ID": "applyforGrantPaperApplication", "Name": "PA1P/PA1A/Solicitors", "Description": "Create application for paper form", "DisplayOrder": 116, "PostConditionState": "CasePrinted", "CallBackURLAboutToSubmitEvent": "http://${CCD_DEF_CASE_SERVICE_BASE_URL}/case/paperForm", "SecurityClassification": "Public", "ShowSummary": "Y", "ShowEventNotes": "Y"},
+  {"LiveFrom": "01/01/2017", "CaseTypeID": "GrantOfRepresentation", "ID": "applyforGrantPaperApplication", "Name": "PA1P/PA1A/Solicitors", "Description": "Create application for paper form", "DisplayOrder": 116, "PostConditionState": "CasePrinted", "CallBackURLAboutToSubmitEvent": "http://${CCD_DEF_CASE_SERVICE_BASE_URL}/case/paperForm", "SecurityClassification": "Public", "ShowSummary": "Y", "ShowEventNotes": "Y", "Publish": "${CCD_DEF_PUBLISH}"},
```
raghera: Did you also test the CaseEventToFields.json files?

raghera: Is this event one where we will actually want to publish in future?  I.e. one that's in one of the ICE stories?

ccdImports/conversionScripts/createAllXLS.sh
raghera: Also I think you are always overwriting this later anyway so seems like it can be set internally anyway.

ccdImports/conversionScripts/convertJsonToXLS.sh
```diff
@@ -27,5 +27,6 @@ docker run --rm --name json2xlsx \
     -v ${ccd_definition_excel_output_file}:/tmp/ccd-definition.xlsx \
     -e CCD_DEF_CASE_SERVICE_BASE_URL \
     -e CCD_DEF_AAC_URL=${CCD_DEF_AAC_URL} \
+    -e CCD_DEF_PUBLISH=${CCD_DEF_PUBLISH} \
```
raghera: Should give this a default in case it is ever unset  
E.g.   `-e CCD_DEF_PUBLISH=${CCD_DEF_PUBLISH:-N}`

Sam applies to the -pipeline too.

raghera (commented):
Looks good generally Scott but made a couple of comments. Take a look.

ccdImports/conversionScripts/convertJsonToXLS.sh
scottiee-jc: I'll add this now

ccdImports/conversionScripts/createAllXLS.sh
scottiee-jc: I think I added it initially and forgot to remove it when I was playing around with the env variables but yes you're right, its overwritten in lines 36-39. Have removed now!

ccdImports/configFiles/CCD_Probate_Backoffice/CaseEvent.json
scottiee-jc: I believe this is one we discussed on call - have crossed checked with the stories we have set up for the ICE epic and it features there

scottiee-jc: I did not actually - let me add a field there also

raghera: Ok.  We may have to discuss if this is ok to merge into the `master` branch.  Should be but we'd want to be sure as this is an existing prod event.
`Publish` being absent and set to `N` behaves the same.

FeliTam (approved):

mikebrownccd (approved):
LGTM

raghera (approved):

PR #3854, opened 2026-05-20
DTSPB-5609 upgrade Elasticsearch v9

FeliTam:
DTSPB-5609

### Change description ###
upgrade Elasticsearch v9

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/test/java/uk/gov/hmcts/probate/service/CaseSearchServiceTest.java
+++ b/src/test/java/uk/gov/hmcts/probate/service/CaseSearchServiceTest.java
@@ -1,5 +1,6 @@
 package uk.gov.hmcts.probate.service;
 
+import com.jayway.jsonpath.JsonPath;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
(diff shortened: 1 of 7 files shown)


src/test/java/uk/gov/hmcts/probate/service/CaseSearchServiceTest.java
```diff
@@ -123,12 +129,54 @@ void findCaseByRecordId() {
...
+        ArgumentCaptor<String> queryCaptor = ArgumentCaptor.forClass(String.class);
+
+        verify(elasticSearchService).runQuery(
+                eq(GRANT_OF_REPRESENTATION),
+                queryCaptor.capture()
+        );
+
+        String query = queryCaptor.getValue();
+
+        assertAll(
```
tom-saunders-cts, line 163: i do wonder whether we should be using a json matcher rather than just expecting the resulting stringified output to contain exactly these values

tom-saunders-cts (dismissed):
i think in terms of the change requested this meets what's been asked - but i'm less confident we're setting a good pattern for some of the testing (though we are just being consistent to what exists?)

src/test/java/uk/gov/hmcts/probate/service/CaseSearchServiceTest.java
FeliTam: updated the test

MarcusHMCTS (approved):

PR #3873, opened 2026-05-28
DTSPB-5573 notification for cases in the state BORedecNotificationSent

MarcusHMCTS:

[DTSPB-5573](https://tools.hmcts.net/jira/browse/DTSPB-5573)

### Change description ###

notification for cases in the state BORedecNotificationSent

```
[ ] Yes
[x] No
```


FeliTam (dismissed):
The changes look sensible to me and are similar to the current implementation for stop reminders.
I can see that firstRedecReminderSentDate is reset in defaultRedeclarationSOTValues. If a superuser uses the change state event to move a case between stop and redeclaration states, that scenario would not be covered. However, I believe we generally ignore that event anyway.

FeliTam (approved):

PR #3889, opened 2026-06-09
DTSPB-5668 Use scanning date as application submitted date for bulkscan case

FeliTam:
[DTSPB-5668](https://tools.hmcts.net/jira/browse/DTSPB-5668)

### Change description ###
Use scanning date as application submitted date for bulkscan case

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/probate/model/exceptionrecord/ExceptionRecordRequest.java
```diff
@@ -315,4 +317,12 @@ public ExceptionRecordOCRFields getOCRFieldsObject() {
...
             .build();
     }
+
+    public Optional<LocalDate> getFormScannedDate() {
+        return scannedDocuments.stream()
+                .filter(doc -> "form".equalsIgnoreCase(doc.type))
+                .findFirst()
+                .map(doc -> doc.scannedDate)
+                .map(LocalDateTime::toLocalDate);
+    }
```
tom-saunders-cts, line 327: is it reasonable to assume that all documents have the same date (or should we find all dates and use the first/last?)

src/main/java/uk/gov/hmcts/probate/service/exceptionrecord/ExceptionRecordService.java
```diff
@@ -106,7 +106,9 @@ public SuccessfulTransformationResponse createCaveatCaseFromExceptionRecord(
                 .map(it -> documentMapper.toCaseDoc(it, erRequest.getExceptionRecordId()))
                 .toList());
 
-            caveatData.setApplicationSubmittedDate(erRequest.getDeliveryDate().toLocalDate());
+            LocalDate scannedDate = erRequest.getFormScannedDate()
+                    .orElseThrow(() -> new OCRMappingException("Unable to map scanned date from scanned document"));
+            caveatData.setApplicationSubmittedDate(scannedDate);
```
tom-saunders-cts, line 111: do we want to throw here or should we define a fallback option ? (if we throw then are we making more work for a caseworker?)

MarcusHMCTS (dismissed):

PR #3901, opened 2026-06-12
DTSPB-5707 - 1A Awaiting Documentation: Examine Digital Case Probate: PA1P/PA1A/Solicitors Manual 

scottiee-jc:

https://tools.hmcts.net/jira/browse/DTSPB-5707

### Change description ###

1A Awaiting Documentation: Examine Digital Case - Probate

```
[ ] Yes
[x] No
```


src/main/resources/dmn/wa-task-initiation-probate-grantofrepresentation.dmn
```diff
@@ -165,6 +165,35 @@ count(list: additionalData.Data.boHandoffReasonList) &gt; 0) then
...
+          <text>"applyforGrantPaperApplicationMan"</text>
+        </inputEntry>
+        <inputEntry id="UnaryTests_0a3bnkb">
+          <text>"CasePrinted"</text>
+        </inputEntry>
+        <inputEntry id="LiteralExpression_1whdf9l">
+          <text>false</text>
+        </inputEntry>
+        <inputEntry id="UnaryTests_1qrdf9q">
+          <text></text>
```
raghera: Does this apply to all caseTypes or just GOR as there is no restriction?

src/main/resources/dmn/wa-task-permissions-probate-grantofrepresentation.dmn
```diff
@@ -72,7 +72,7 @@
           <text>"CTSC"</text>
         </outputEntry>
         <outputEntry id="LiteralExpression_0nlp78t">
-          <text></text>
+          <text>"SKILL:ABA6:ProbateExamining"</text>
```
raghera, line 75: I couldn't see in the story if this was called out.  It may make sense but just checking it's the intention to require team-leaders to have the skill to get the team leader permissions.

src/main/resources/dmn/wa-task-permissions-probate-grantofrepresentation.dmn
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<definitions xmlns="https://www.omg.org/spec/DMN/20191111/MODEL/" xmlns:dmndi="https://www.omg.org/spec/DMN/20191111/DMNDI/" xmlns:dc="http://www.omg.org/spec/DMN/20180521/DC/" xmlns:biodi="http://bpmn.io/schema/dmn/biodi/2.0" xmlns:camunda="http://camunda.org/schema/1.0/dmn" id="wa-permissions-definition" name="DRD" namespace="http://camunda.org/schema/1.0/dmn" exporter="Camunda Modeler" exporterVersion="5.33.1">
+<definitions xmlns="https://www.omg.org/spec/DMN/20191111/MODEL/" xmlns:dmndi="https://www.omg.org/spec/DMN/20191111/DMNDI/" xmlns:dc="http://www.omg.org/spec/DMN/20180521/DC/" xmlns:biodi="http://bpmn.io/schema/dmn/biodi/2.0" xmlns:camunda="http://camunda.org/schema/1.0/dmn" id="wa-permissions-definition" name="DRD" namespace="http://camunda.org/schema/1.0/dmn" exporter="Camunda Modeler" exporterVersion="5.46.1">
```
raghera, line 2: I assume no changes are required for the configuration DMN, is that correct?  
Did you check if you needed to add anything to the Task Types DMN too?

src/main/resources/dmn/wa-task-cancellation-probate-grantofrepresentation.dmn
```diff
@@ -24,13 +24,13 @@
...
         <inputEntry id="UnaryTests_1n3g5ie">
-          <text></text>
+          <text>"CasePrinted"</text>
         </inputEntry>
         <inputEntry id="UnaryTests_1g1cc7g">
           <text>"boWithdrawApplicationForCasePrinted"</text>
         </inputEntry>
         <inputEntry id="UnaryTests_1ah5t17">
-          <text></text>
+          <text>"BOCaseClosed"</text>
```
raghera, line 33: I don't think it is in the story so can leave it for now but do we want to specify that all Tasks for this Case would be Cancelled when this event happens?

During Initiation we set the processCategoryIdentifier as "case progression" which you could then use during cancellation to target only Tasks that have been created with that identifier.

See: https://tools.hmcts.net/confluence/pages/viewpage.action?pageId=1915168142#HLDTaskManagementv1.6-2.12.5Configuration-DrivenProcessCancellation

src/test/java/uk/gov/hmcts/probate/dmn/CamundaTaskWaCancellationTest.java
```diff
@@ -29,7 +29,10 @@ public static void initialization() {
 
     public static Stream<Arguments> scenarioProvider() {
         return Stream.of(
+                // Withdraw application from CasePrinted ("Awaiting documentation") cancels all tasks on the case
                 event("boWithdrawApplicationForCasePrinted")
+                        .fromState("CasePrinted")
+                        .toState("BOCaseClosed")
```
raghera: Since we are proposing on relying on DMN tests it's worthwhile covering some negative cases too.  E.g. where nothing should be cancelled because the from/to state is empty or the event is different.

This probably applies to all DMN tests and could be something to discuss as a team.

raghera (commented):
Looks good Scott but take a look at my comments

src/main/resources/dmn/wa-task-initiation-probate-grantofrepresentation.dmn
scottiee-jc: Previously i saw an issue where the task wasn't being created when specifying the caseType - connected with Mike who explained there were some authorisation case event to field entries that needed to be added for this to take effect properly. Having added those, it is working now! So, I have added "gop" to this case type

src/main/resources/dmn/wa-task-permissions-probate-grantofrepresentation.dmn
scottiee-jc: I think this makes sense here - for this specific permission it is for ExamineDigitalCaseProbate, and the team leader will need this skill to view the tasks available to them

src/main/resources/dmn/wa-task-permissions-probate-grantofrepresentation.dmn
scottiee-jc: None that I can see are relevant for the configuration dmn for this one. Unsure on the task types dmn, might need guidance on that one

src/test/java/uk/gov/hmcts/probate/dmn/CamundaTaskWaCancellationTest.java
scottiee-jc: This sounds like a good idea, I will try and write some

scottiee-jc: Have now finished writing these @raghera

src/main/resources/dmn/wa-task-cancellation-probate-grantofrepresentation.dmn
scottiee-jc: I have updated this with the output clause for 'case progression' for this specific rule. For warning code/ text, what are the expected values?

raghera: Warning it's only relevant if your action output is `WARN`.  Basically to update an existing Task with a warning marker.

src/main/resources/dmn/wa-task-permissions-probate-grantofrepresentation.dmn
raghera: Tasks types DMN is maintaining a list of Task Types available.  I think it's worthwhile keeping it updated as we go along in Case we miss any.  Mike's PR has added an entry so check if that covers your scenario or if you need to add another line?

src/main/resources/dmn/wa-task-cancellation-probate-grantofrepresentation.dmn
scottiee-jc: I see, will keep a note of that

src/test/java/uk/gov/hmcts/probate/dmn/CamundaTaskWaCancellationTest.java
```diff
@@ -45,18 +38,59 @@ void if_this_test_fails_needs_updating_with_your_changes() {
...
+                        cancellationProperties.get("processCategories"));
+            }
+        }
+        return inputVariables;
     }
 
+    private void testBoWithdrawApplicationEvent(List<Map<String, Object>> dmnResultList,
+                                                Map<String, String> cancellationProperties) {
+        if (cancellationProperties.containsValue("CasePrinted")
+                || cancellationProperties.containsValue("BOCaseClosed")) {
```
raghera: Can we check this is correct.  From your DMN fromState must be `CasePrinted` AND toState must be `BOCaseClosed` but the above has an `||` which doesn't look correct. 
Obviously it has to be the correct event too.

src/test/java/uk/gov/hmcts/probate/dmn/CamundaTaskWaCancellationTest.java
```diff
@@ -45,18 +38,59 @@ void if_this_test_fails_needs_updating_with_your_changes() {
...
-        inputVariables.putValue("state", state);
+        inputVariables = putAllCancellationProperties(inputVariables, cancellationProperties);
 
         DmnDecisionTableResult dmnDecisionTableResult = evaluateDmnTable(inputVariables);
-        assertThat(dmnDecisionTableResult.getResultList(), is(expectedDmnOutcome));
+        List<Map<String, Object>> dmnResultList = dmnDecisionTableResult.getResultList();
+
+        // no action key means the no results will be returned
+        if (!cancellationProperties.containsKey("action")) {
+            Assertions.assertEquals(0, dmnResultList.size());
```
raghera: Bit unclear why this assertion is required here?

src/test/java/uk/gov/hmcts/probate/dmn/CamundaTaskWaCancellationTest.java
raghera: Can see you've added some more tests so will resolve this comment.

src/main/resources/dmn/wa-task-initiation-probate-grantofrepresentation.dmn
raghera: Can see the update thanks.

src/test/java/uk/gov/hmcts/probate/dmn/CamundaTaskWaCancellationTest.java
```diff
@@ -45,18 +38,59 @@ void if_this_test_fails_needs_updating_with_your_changes() {
...
+            inputVariables.putValue("fromState", cancellationProperties.get("fromState"));
+            inputVariables.putValue("state", cancellationProperties.get("state"));
+            if (cancellationProperties.size() == 3) {
+                return inputVariables;
+            }
+            inputVariables.putValue("action", cancellationProperties.get("action"));
+            // For when it is a 'cancel all' type of task cancellation
+            if (cancellationProperties.containsKey("processCategories")) {
+                inputVariables.putValue("processCategories",
+                        cancellationProperties.get("processCategories"));
```
raghera, line 72: I think you can leave it for now but put it here for future reference.  
You can probably refactor this with a loop?  Keep a list of input keys and add to your inputVariable list.  This way if you add any in future it would still work.

raghera (commented):
Think you've addressed most but I added a few more around the Cancellation test.

src/test/java/uk/gov/hmcts/probate/dmn/CamundaTaskWaCancellationTest.java
scottiee-jc: Ah, think i forgot to remove this after playing around with the testing yesterday. Heat was getting to me I think!

src/test/java/uk/gov/hmcts/probate/dmn/CamundaTaskWaCancellationTest.java
scottiee-jc: I'll make a note of this for future test cases.

src/test/java/uk/gov/hmcts/probate/dmn/CamundaTaskWaCancellationTest.java
scottiee-jc: You're correct, have changed this now

Jenkinsfile_CNP
FeliTam, line 1: Please revert the changes in Jenkinsfile_CNP before merging.
We have separate tech debt ticket in place for the migration. Thanks.

FeliTam (changes requested):
The changes look good to me, apart from the Jenkinsfile. Thanks!

Jenkinsfile_CNP
scottiee-jc: @FeliTam Resolved

FeliTam (dismissed):
LGTM

raghera (approved):

PR #3911, opened 2026-06-19
DTSPB-5113 reimpl

FeliTam:
DTSPB-5113

### Change description ###
Migrate dormant case to previous (post-grant) state if issued

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/probate/service/ccd/AuditEventService.java
+++ b/src/main/java/uk/gov/hmcts/probate/service/ccd/AuditEventService.java
@@ -50,4 +50,24 @@ public class AuditEventService {
                 .skip(1)
                 .findFirst();
     }
+
+    public Optional<AuditEvent> getLatestAuditEventExcludingDormantState(String caseId, List<String> stateNames,
+                                                                         String userToken, String authToken) {
+        String maxEventId = caseDataApi.getAuditEvents(userToken, authToken, false, caseId)
+                .getAuditEvents().stream()
+                .max(Comparator.comparing(AuditEvent::getId))
+                .map(AuditEvent::getId)
+                .orElse(null);
+        if (maxEventId == null) {
+            return Optional.empty();
+        }
+        log.info("Getting latest audit event for caseId: {}", caseId);
+        AuditEventsResponse auditEventsResponse
+                = caseDataApi.getAuditEvents(userToken, authToken, false, caseId);
+        log.info("auditEventsResponse AuditEvents().size(): {}", auditEventsResponse.getAuditEvents().size());
+        return auditEventsResponse.getAuditEvents().stream()
+                .filter(auditEvent -> stateNames.contains(auditEvent.getStateId())
+                        && auditEvent.getId().equals(maxEventId))
+                .findFirst();
+    }
 }
--- a/src/main/java/uk/gov/hmcts/probate/service/ccd/AuditEventService.java
+++ b/src/main/java/uk/gov/hmcts/probate/service/ccd/AuditEventService.java
@@ -50,24 +50,4 @@ public class AuditEventService {
                 .skip(1)
                 .findFirst();
     }
-
-    public Optional<AuditEvent> getLatestAuditEventExcludingDormantState(String caseId, List<String> stateNames,
-                                                                         String userToken, String authToken) {
-        String maxEventId = caseDataApi.getAuditEvents(userToken, authToken, false, caseId)
-                .getAuditEvents().stream()
-                .max(Comparator.comparing(AuditEvent::getId))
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/hmcts/probate/service/ccd/AuditEventService.java
```diff
@@ -50,4 +50,24 @@ public Optional<AuditEvent> getPreviousAuditEventOfByEventId(String caseId, Even
...
+            return Optional.empty();
+        }
+        log.info("Getting latest audit event for caseId: {}", caseId);
+        AuditEventsResponse auditEventsResponse
+                = caseDataApi.getAuditEvents(userToken, authToken, false, caseId);
+        log.info("auditEventsResponse AuditEvents().size(): {}", auditEventsResponse.getAuditEvents().size());
+        return auditEventsResponse.getAuditEvents().stream()
+                .filter(auditEvent -> stateNames.contains(auditEvent.getStateId())
+                        && auditEvent.getId().equals(maxEventId))
+                .findFirst();
```
tom-saunders-cts: this is a very picky question but is there a guarantee that the audit events are ordered most recent to least? (it seems like a good assumption but...)

src/main/java/uk/gov/hmcts/probate/service/ccd/AuditEventService.java
```diff
@@ -50,4 +50,24 @@ public Optional<AuditEvent> getPreviousAuditEventOfByEventId(String caseId, Even
                 .skip(1)
                 .findFirst();
     }
+
+    public Optional<AuditEvent> getLatestAuditEventExcludingDormantState(String caseId, List<String> stateNames,
```
tom-saunders-cts: maybe this should be `...InStateList` rather than `...ExcludingDormantState` since the behaviour is reliant on the `stateNames` parameter rather than just not dormant

src/main/java/uk/gov/hmcts/probate/service/migration/Dtspb5113GorMigration.java
```diff
@@ -0,0 +1,59 @@
...
+import uk.gov.hmcts.probate.security.SecurityUtils;
+import uk.gov.hmcts.probate.service.ccd.AuditEventService;
+
+import java.util.List;
+
+
+
+@Slf4j
+@Component
+public class Dtspb5113GorMigration implements GorMigrationHandler {
```
tom-saunders-cts, line 14: i sort of wonder whether the action to find the prior state might be 'cleaner' in the migration tool, and then we could just use a semi-generic handler which takes the target state as an input (which we could use for both migration/rollback here _and_ any other ticket where all we're interested in is changing the state)

this would also reduce the complexity/overhead of the callback which might be beneficial

src/main/java/uk/gov/hmcts/probate/service/ccd/AuditEventService.java
FeliTam: updated

src/main/java/uk/gov/hmcts/probate/service/migration/Dtspb5113GorMigration.java
FeliTam: moved the function in AuditEventService to the migration tool

src/main/java/uk/gov/hmcts/probate/service/ccd/AuditEventService.java
FeliTam: added sorting and moved to migration tool

src/test/java/uk/gov/hmcts/probate/service/migration/Dtspb5113GorMigrationTest.java
```diff
@@ -0,0 +1,85 @@
...
+@ExtendWith(MockitoExtension.class)
+class Dtspb5113GorMigrationTest {
+
+    private Dtspb5113GorMigration underTest;
+
+    @Mock
+    SecurityUtils securityUtils;
+
+    @Mock
+    AuditEventService auditEventService;
```
tom-saunders-cts: i don't think this is actually needed any more?

src/test/java/uk/gov/hmcts/probate/service/ccd/AuditEventServiceTest.java
tom-saunders-cts, line 1: we're not actually making any change here other than the annotation?

tom-saunders-cts (dismissed):
there's a couple of minor nitpicks i the comments but nothing which should meaningfully delay this afaict

jamiealbertelli (approved):

PR #3937, opened 2026-06-30
DTSPB-5680: Stop documents being replaced

jamiealbertelli:
[DTSPB-5680](https://tools.hmcts.net/jira/browse/DTSPB-5680)

### Change description ###
Stop caseworkers from being able to remove/replace uploaded docs from the 'Upload documents' event

```
[ ] Yes
[ ] No
```


src/main/java/uk/gov/hmcts/probate/service/EvidenceUploadService.java
```diff
@@ -3,19 +3,77 @@
...
+
+        Map<String, UploadDocument> afterById = new HashMap<>();
+        for (CollectionMember<UploadDocument> document : documentsAfter) {
+            if (document != null && document.getId() != null) {
+                afterById.put(document.getId(), document.getValue());
+            }
+        }
+
+        for (CollectionMember<UploadDocument> existingDocument : documentsBefore) {
+            if (existingDocument == null || existingDocument.getId() == null) {
```
MarcusHMCTS: existingDocument.getId() with many times
extract it into a local variable?

src/main/java/uk/gov/hmcts/probate/service/EvidenceUploadService.java
```diff
@@ -3,19 +3,77 @@
...
+                continue;
+            }
+            UploadDocument afterDocument = afterById.get(existingDocument.getId());
+            if (!afterById.containsKey(existingDocument.getId())
```
MarcusHMCTS: afterById.get & afterById.containsKey duplicate check
can afterDocument!=null instead?

MarcusHMCTS (dismissed):

src/main/java/uk/gov/hmcts/probate/service/EvidenceUploadService.java
```diff
@@ -3,19 +3,77 @@
...
+            boolean hasAfterDocument = afterById.containsKey(existingDocumentId);
+            UploadDocument afterDocument = hasAfterDocument ? afterById.get(existingDocumentId) : null;
+            if (!hasAfterDocument
+                    || !Objects.equals(existingDocument.getValue(), afterDocument)) {
+                Long caseId = Optional.ofNullable(callbackRequest.getCaseDetails())
+                        .map(CaseDetails::getId)
+                        .orElse(null);
+                log.error("Document with ID {} has been modified or removed for case {}",
+                        existingDocumentId,
+                        caseId);
```
FeliTam: this is comparing the id from CollectionMember rather than the actual document id (document url)
CollectionMember id may change during the event for some reason, even though the same document has not been updated.

src/main/java/uk/gov/hmcts/probate/service/EvidenceUploadService.java
```diff
@@ -3,19 +3,77 @@
...
+                continue;
+            }
+            String existingDocumentId = existingDocument.getId();
+            boolean hasAfterDocument = afterById.containsKey(existingDocumentId);
+            UploadDocument afterDocument = hasAfterDocument ? afterById.get(existingDocumentId) : null;
+            if (!hasAfterDocument
+                    || !Objects.equals(existingDocument.getValue(), afterDocument)) {
+                Long caseId = Optional.ofNullable(callbackRequest.getCaseDetails())
+                        .map(CaseDetails::getId)
+                        .orElse(null);
```
FeliTam: can just get the case id from CaseDetails without wrapping with Optional

src/main/java/uk/gov/hmcts/probate/service/EvidenceUploadService.java
```diff
@@ -3,19 +3,77 @@
...
+
+        for (CollectionMember<UploadDocument> existingDocument : documentsBefore) {
+            if (existingDocument == null || existingDocument.getId() == null) {
+                continue;
+            }
+            String existingDocumentId = existingDocument.getId();
+            boolean hasAfterDocument = afterById.containsKey(existingDocumentId);
+            UploadDocument afterDocument = hasAfterDocument ? afterById.get(existingDocumentId) : null;
+            if (!hasAfterDocument
+                    || !Objects.equals(existingDocument.getValue(), afterDocument)) {
```
FeliTam: I don't think we can compare the UploadDocument object directly, as that would prevent caseworkers from updating the comment on the document.

FeliTam (changes requested):

src/test/java/uk/gov/hmcts/probate/service/EvidenceUploadServiceTest.java
```diff
@@ -40,5 +49,119 @@ void shouldVerifyUpdateOfLastEvidenceAddedDate() {
...
+        CaseDetails beforeDetails = new CaseDetails(beforeData, null, 123L);
+        CaseDetails afterDetails = new CaseDetails(afterData, null, 123L);
+        CallbackRequest callbackRequest = new CallbackRequest(afterDetails);
+        callbackRequest.setCaseDetailsBefore(beforeDetails);
+
+        assertDoesNotThrow(() -> evidenceUploadService.validateExistingUploadedDocuments(callbackRequest));
+    }
+
+    @Test
+    void shouldThrowNullPointerExceptionWhenCaseDetailsAfterIsNull() {
```
FeliTam: think we shouldn't expect NullPointerException. The controller only catches BusinessValidationException, so this becomes a 500.
can we resolve the case ID safely and retain the validation error.

src/main/java/uk/gov/hmcts/probate/service/EvidenceUploadService.java
```diff
@@ -3,19 +3,82 @@
...
+                .orElse(Collections.emptyList());
+
+        Set<String> afterDocumentUrls = documentsAfter.stream()
+                .map(this::getDocumentUrl)
+                .filter(Objects::nonNull)
+                .collect(Collectors.toSet());
+
+        for (CollectionMember<UploadDocument> existingDocument : documentsBefore) {
+            String existingDocumentUrl = getDocumentUrl(existingDocument);
+            if (existingDocumentUrl == null) {
```
FeliTam, line 73: Skipping an existing document with no URL is fail-open. should we reject it instead?

src/test/java/uk/gov/hmcts/probate/service/EvidenceUploadServiceTest.java
FeliTam, line 1: can we add the happy-path test: existing document remains unchanged and a new document is added

FeliTam (dismissed):

FeliTam (approved):

PR #4112, opened 2026-08-18
fix: add authToken parameter to refundsApprovalLiberata

FeliTam:
### Change description ###
fix: add authToken parameter to refundsApprovalLiberata

```
[ ] Yes
[x] No
```


src/test/PlaywrightTest/Pages/IDAM/signIn.ts
tom-saunders-cts: i'm going to trust that the repeated replacement of the old explicit login handling to the single implementation doesn't lose any special handling in one of them hidden inside another call

tom-saunders-cts (approved):

